PR #40, opened 2017-05-30
Styling example

tobrun:
This PR is a proposition for changing the public API for LocationLayer from setters to using a style.xml format. This xml is provided at startup and users can update the styling with something as `LocationLayerPlugin#applyStyle(@StyleRes int styleRes)`.

This PR currently only exposes support for BackgroundDrawable. When we reach consensus on how we are exposing the API, I will refactor the other implementations in as well. 

cc @cammace  @zugaldia


cammace:
Going ahead and merging this PR into the original one to finish up this feature.

PR #44, opened 2017-06-02
add tests for traffic

tobrun:
As first official plugin to this repo, we need to set a good example for others and add tests!
  - Unit tests for traffic module 
  - Instrumentation tests for the test app (we need an activity to run these)
  - dependencies.gradle setup cfr mapbox-java/mapbox-gl-native

What does it test:
 - if mapboxmap and traffic plugin are correctly created
 - If traffic source is added on init
 - If traffic layers are added on init
 - if traffic layers are visible on init
 - if traffic layers are invisble after toggling the plugin
 - if traffic min zoom level = expected value
 - loading a new style should readd the source/layers
 - lineWidth
 - filter

### DIFF
--- a/plugins/app/build.gradle
+++ b/plugins/app/build.gradle
@@ -1,42 +1,69 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion "25.0.2"
+    compileSdkVersion rootProject.ext.compileSdkVersion
+    buildToolsVersion rootProject.ext.buildToolsVersion
+
     defaultConfig {
         applicationId "com.mapbox.mapboxsdk.plugins.testapp"
-        minSdkVersion 15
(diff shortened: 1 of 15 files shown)


tobrun:
Currently still a WIP but feel free to 👀 already

tobrun:
Reworked the test to test one thing at a time. Two items remain to be tested:
 - lineWidth functions (not working as I would expect)
 - filter (requires getFilter to be implemented)

Feel free to review

tobrun:
All tests that have been added, can I get a review @Guardiola31337 @zugaldia ?

tobrun:
32c5908 fixes up naming of tests without the "test" prefix and actually executes the instrumentation test on firebase. Can I get a review on this?

tobrun:
friendly bump on a review @Guardiola31337 & @zugaldia

Issue #4, opened 2017-04-05
Traffic plugin


zugaldia:
We'd like to add a traffic plugin that lets you add a traffic layer to any Mapbox basemap. To do that, we could use runtime styling.

We'd also like to add a sample activity to the test app that lets you switch between a few basemaps, including a toggle to turn traffic on and off.

cc: @ericrwolfe

Issue #15, opened 2017-04-15
Kotlin plugin


zugaldia:
This ticket tracks an idea about the _language_ to write the plugin with rather than the actual purpose of the plugin. Considering plugins are well-contained, a plugin would be an ideal place to showcase how to interact with the Maps SDK using only Kotlin.

What do you think @mapbox/android ?

tobrun:
One idea I have been having is using Kotlin extension functions to add functionality to already existing classes. Eg. you can add dynamically add functions to MapboxMap based on the plugins you are pulling in.

zugaldia:
The more I think about this the more I'm convinced that one good path here would be to build a DSL around the map. What if we could do something like this?

```
map {
	annotations {
		for (marker in markers)
			marker { +marker.location }
	}

	sources {
		...
	}

	layers {
		...
	}
}
```

Similar to what Anko does for Android layouts (https://github.com/Kotlin/anko#anko-layouts-wiki).

cammace:
Stale conversation and Kotlin discussion have been being made in gl-internal. Closing

PR #46, opened 2017-06-11
Kotlin Traffic plugin PoC

Guardiola31337:
In order to experiment with Kotlin I thought that translating the [Traffic Plugin](https://github.com/mapbox/mapbox-plugins-android/issues/4) into Kotlin would be a great proof of concept.

Here are my initial findings:

- `Convert Java File to Kotlin File` AS option doesn't work that well... Especially with complex classes. I tried to convert `TrafficPlugin.java` and Kotlin generated class was full of errors, so I decided to do it step by step porting each class/method slowly but surely (learning all pitfalls along the way).
- I found a bunch of problems dealing with type safety of Java Generics (e.g. [`Function.java`](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/style/functions/Function.java)) because Kotlin compiler is much stricter.
- [Varargs in Kotlin](https://kotlinlang.org/docs/reference/functions.html#variable-number-of-arguments-varargs) are a bit different. You need to use the _spread_ operator (prefix the array with `*`) to indicate that it's a vararg.

Next steps:

- [ ] [Calling Kotlin from Java](https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#calling-kotlin-from-java)
  - [ ] Understand Kotlin-related annotations (i.e. `@JvmName`, `@JvmField`, etc.)
- [ ] Take advantage of FP and Kotlin features (e.g. extension functions) to build well-designed DSLs across our SDKs
- [ ] [Documenting Kotlin Code](https://kotlinlang.org/docs/reference/kotlin-doc.html#documenting-kotlin-code) (_KDoc_)
- [ ] Analyze
  - [ ] [klint](https://ktlint.github.io/)
  - [ ] [detekt](https://github.com/arturbosch/detekt)
- [ ] Continue exploring and experimenting
- [ ] Add your topic here

cc/ @zugaldia


cammace:
What's the status on this plugin or was this just an experiment to learn Kotlin? The traffic plugin already exists so I'm not sure what the difference between this plugin and the one already released is.

Guardiola31337:
@cammace 
> What's the status on this plugin or was this just an experiment to learn Kotlin?

It was an experiment which its main purpose was to start exploring how to build plugins completely written in Kotlin (analyzing pros and cons, pitfalls, etc.) and to start thinking about using the language across our SDKs (examining what it offers us in terms of implementing better libraries).

> The traffic plugin already exists so I'm not sure what the difference between this plugin and the one already released is.

Actually, there is no difference. No need to get merged.
But it could be useful though to compare a plugin written in Java vs one written in Kotlin and take it as a case of study.

What do you think?

zugaldia:
> Actually, there is no difference. No need to get merged.
But it could be useful though to compare a plugin written in Java vs one written in Kotlin and take it as a case of study.

Before we close this PR (agreed on not needing to merge it) I'd like to save any learnings from this process while we keep researching https://github.com/mapbox/mapbox-plugins-android/issues/15. @Guardiola31337 How about we move the most relevant pieces of this PR to a wiki page in this repo? It could be a resource that we share with developers starting to use our SDKs with Kotlin.

cammace:
I think this PR is safe to close now? Otherwise I can rebase since we have adjusted the file structure in this project.

tobrun:
stale

Issue #60, opened 2017-07-17
Only add traffic layer above if layer exist


cammace:
One issue restricting the traffic plugin from working on any platform is the fact that we are trying to add the traffic above `"bridge-motorway"` but this layer id doesn't exist in every style causing the traffic layers to not get added. 

cc: @tobrun @zugaldia

Issue #55, opened 2017-07-14
Replace traffic.toggle() with setVisibility(boolean)


cammace:
To keep consistency across plugins we should change the traffic plugin to have `setVisibility` rather  than `toggle` so it's easier on the developer to, for example, directly pass in a boolean setting into the params.

cc: @tobrun @danesfeder

Issue #14, opened 2017-04-11
Traffic Plugin add above road layer 


tobrun:
We added a Traffic Plugin in #6, currently we hardcode the `layerId` which we use in `addLayerBefore(layer, layedId)`. Ideally we use can use the source layer identifier to check for a `road`  identifier instead. 

Upstream feature ticket for source layer identifier in https://github.com/mapbox/mapbox-gl-native/issues/8663

tobrun:
Adding here that we need to use source layer identifier in setVisibility as well.

PR #61, opened 2017-07-17
Few improvements to traffic plugin

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/60
Closes https://github.com/mapbox/mapbox-plugins-android/issues/55
Closes https://github.com/mapbox/mapbox-plugins-android/issues/14?

I've gone ahead and made a few changes to the traffic plugin to keep consistency across plugins and increase compatibility to all Mapbox map styles. Only issue we should resolve (@tobrun mentioned there's a better way to do this) is to change `if (!(styleLayers.get(i) instanceof SymbolLayer))` into something more performant.

### DIFF
--- a/plugins/traffic/src/main/java/com/mapbox/mapboxsdk/plugins/traffic/TrafficPlugin.java
+++ b/plugins/traffic/src/main/java/com/mapbox/mapboxsdk/plugins/traffic/TrafficPlugin.java
@@ -188,16 +188,21 @@ public final class TrafficPlugin implements MapView.OnMapChangedListener {
       Local.FUNCTION_LINE_OPACITY_CASE
     );
 
-    // #TODO https://github.com/mapbox/mapbox-plugins-android/issues/14
     addTrafficLayersToMap(localCase, local, placeLayerBelow());
   }
 
+  /**
+   * Attempts to find the layer which the traffic should be placed below. Depending on the style, this might not always
+   * be accurate.
+   */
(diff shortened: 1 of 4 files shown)


plugins/traffic/src/main/java/com/mapbox/mapboxsdk/plugins/traffic/TrafficPlugin.java
```diff
@@ -169,7 +189,19 @@ private void addLocalLayer() {
...
+  private String placeLayerBelow() {
+    if (belowLayer == null || belowLayer.isEmpty()) {
+      List<Layer> styleLayers = mapboxMap.getLayers();
+      for (int i = styleLayers.size() - 1; i >= 0; i--) {
+        if (!(styleLayers.get(i) instanceof SymbolLayer)) {
+          return styleLayers.get(i).getId();
+        }
+      }
+    }
+    return belowLayer;
```
tobrun: +  private String getIdLayerBelow() {
+    if (belowLayer == null || belowLayer.isEmpty()) {
+      List<Layer> styleLayers = mapboxMap.getLayers();
        Layer layer;
+      for (int i = styleLayers.size() - 1; i >= 0; i--) {
          layer = styleLayers.get(i);
+        if (!(layer instanceof SymbolLayer)) {
+          return layer.getId();
+        }
+      }
+    }
+    return belowLayer;

PR #74, opened 2017-08-07
Localization plugin

### LINKED ISSUE #36: Localization plugin
We should have a plugin that automatically localizes map labels into the user’s preferred language. There’s already a robust implementation of this feature in Objective-C targeting the macOS SDK that could be ported to the Android SDK:

https://github.com/mapbox/mapbox-gl-native/blob/cc1ea759a681cadcfd06fd7bdda79ca6deb38c62/platform/macos/app/MapDocument.m#L346-L393
https://github.com/mapbox/mapbox-gl-native/blob/cc1ea759a681cadcfd06fd7bdda79ca6deb38c62/platform/macos/app/MGLVectorSource+MBXAdditions.m

/ref mapbox/mapbox-gl-native#7031
/cc @tobrun @langsmith @ericrwolfe

langsmith:
This plugin resolves #36  and enables a developer to localize the map text to the language set on the Android device.

The plugin also has individual methods for converting the map to each of [the languages that we support](https://www.mapbox.com/help/change-language/#languages-available). Anyways, here it is. 

ezgif com-resize 2

ezgif com-resize

cc @tobrun  @1ec5 @jcsg


langsmith:
And yes, I will create a test class too

langsmith:
FYI, here's why the test is failing 👇 
screen shot 2017-08-09 at 1 32 29 pm

tobrun:
@langsmith if you aren't adding any instrumentation tests, feel free to remove that example file.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @param mapboxMap the MapboxMap to apply localization plugin with
+   * @since 0.1.0
+   */
+  public LocalizationPlugin(@NonNull final MapboxMap mapboxMap) {
+    listOfMapLayers = mapboxMap.getLayers();
+    String deviceLanguage = Locale.getDefault().getLanguage();
+    for (Layer layer : listOfMapLayers) {
+      // TODO: Fix if() statement? What should if statement be looking for in layer ids?
+      if (layer.getId().contains("")) {
```
cammace: Briefly looking over this, I think you are looking to change all Symbol layers? For that you can do `if(layer instanceof SymbolLayer) {}`

plugins/localization/proguard-rules.pro
```diff
@@ -0,0 +1,25 @@
+# Add project specific ProGuard rules here.
```
cammace: Doesn't look like this ProGuard files needed.

plugins/localization/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">localization</string>
```
cammace: Don't need this file

langsmith: Removed

plugins/localization/proguard-rules.pro
langsmith: Removed

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: Fixed

langsmith:
Note: Cameron's review requests have been addressed.

1ec5:
> I added an individual method for converting the map to each of [the languages that we support](https://www.mapbox.com/help/change-language/#languages-available). Maybe they don't belong in this plugin.

There does need to be a list hard-coded somewhere, until mapbox/tilejson-spec#14 is implemented. (Someday…) Note that the canonical list of available languages is in the [Mapbox Streets source v7 documentation](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#overview).

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Create a localization plugin.
+   *
+   * @param mapboxMap the MapboxMap to apply localization plugin with
+   * @since 0.1.0
+   */
+  public LocalizationPlugin(@NonNull final MapboxMap mapboxMap) {
+    listOfMapLayers = mapboxMap.getLayers();
+    String deviceLanguage = Locale.getDefault().getLanguage();
+    for (Layer layer : listOfMapLayers) {
```
1ec5: What if the style includes layers that use a source other than Mapbox Streets? For example, there may be a layer based on a custom Studio dataset that also has a `{name}` property but no corresponding `{name_*}` properties.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,114 @@
...
+      if (layer instanceof SymbolLayer) {
+        layer.setProperties(PropertyFactory.textField(String.format("{name_%s}", deviceLanguage)));
```
1ec5: What if the symbol layer had no `textField` to begin with, or what if it was set to `{ref}`? Would highway shields start displaying road names instead of route numbers?

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,114 @@
...
+   * Sets map text to Russian.
+   */
+  public void setMapTextToRussian() {
+    setMapTextLanguage("ru");
+  }
+
+  /**
+   * Sets map text to Traditional Chinese.
+   */
+  public void setMapTextToTraditionalChinese() {
```
1ec5: `{name_zh}` is Chinese in general, not necessarily Traditional Chinese. For example, country labels are in Simplified Chinese, as are places in 🇨🇳.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,114 @@
...
+
+  /**
+   * Create a localization plugin.
+   *
+   * @param mapboxMap the MapboxMap to apply localization plugin with
+   * @since 0.1.0
+   */
+  public LocalizationPlugin(@NonNull final MapboxMap mapboxMap) {
+    listOfMapLayers = mapboxMap.getLayers();
+    String deviceLanguage = Locale.getDefault().getLanguage();
```
1ec5: According to [this documentation](https://developer.android.com/reference/java/util/Locale.html), `getLanguage()` returns deprecated ISO 639 codes, such as `iw` for Hebrew instead of the current standard `he`.

Additionally, will this method return `zh-Hans` when the user chooses Chinese (Simplified) in their device settings?

langsmith:
Thanks @1ec5 . Lots more to learn/fix 👍 

>There does need to be a list hard-coded somewhere, until mapbox/tilejson-spec#14 is >implemented. (Someday…) Note that the canonical list of available languages is in the Mapbox >Streets source v7 documentation.

Are there specific changes that I need to make related to ☝️  or was that just a comment? Couldn't tell

1ec5:
>> Note that the canonical list of available languages is in the Mapbox >Streets source v7 documentation.
> 
> Are there specific changes that I need to make related to ☝️  or was that just a comment? Couldn't tell

No changes needed for now. The canonical list is maintained by the cartography team and is more likely to be kept up-to-date than the /help document.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: Cool. Changed the method name to `setMapTextToChinese()` and added explanatory javadocs for both methods. Now reads as 👇 

```
/**
   * Sets map text to Chinese.
   *
   * This method uses simplified Chinese characters for custom label layers: #country_label, #state_label,
   * and #marine_label. All other label layers are sourced from OpenStreetMap and may contain one of several dialects
   * and either simplified or traditional Chinese characters in the {name_zh} field.
   */
  public void setMapTextToChinese() {
    setMapTextLanguage("zh");
  }

  /**
   * Sets map text to Simplified Chinese.
   *
   * Using his method is similar to setMapTextToChinese() (see above), except any Traditional Chinese
   * characters are automatically transformed to Simplified Chinese.
   */
  public void setMapTextToSimplifiedChinese() {
    setMapTextLanguage("zh-Hans");
  }
```

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
```diff
@@ -0,0 +1,121 @@
...
+   * and either simplified or traditional Chinese characters in the {name_zh} field.
+   */
+  public void setMapTextToChinese() {
+    setMapTextLanguage("zh");
+  }
+
+  /**
+   * Sets map text to Simplified Chinese.
+   *
+   * Using his method is similar to setMapTextToChinese() (see above), except any Traditional Chinese
```
1ec5: Typo: `his method`.

1ec5: > (see above)

Doesn’t JavaDoc autolink method references? That would make glosses like this unnecessary.

plugins/app/src/main/res/menu/menu_languages.xml
```diff
@@ -0,0 +1,52 @@
...
+        app:showAsAction="never"/>
+
+    <item
+        android:id="@+id/russian"
+        android:title="Russian"
+        app:showAsAction="never"/>
+
+    <item
+        android:id="@+id/chinese"
+        android:title="Traditional Chinese"
```
1ec5: Replace with “Chinese”.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: Fixed

langsmith: Fixed

plugins/app/src/main/res/menu/menu_languages.xml
langsmith: Fixed

cammace:
I cleaned up a bunch of the code and moved things around to reflect the latest changes made in master. Would be good to continue working on this and merging it.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: https://gist.github.com/SeanZoR/9152167 👀

langsmith: https://developer.android.com/reference/java/util/Locale.html


screen shot 2018-01-04 at 4 26 10 pm

langsmith: https://www.loc.gov/standards/iso639-2/php/code_changes.php

langsmith: https://web.archive.org/web/20131209133616/http://code.google.com/p/android/issues/detail?id=3639

langsmith: I made [this commit](https://github.com/mapbox/mapbox-plugins-android/pull/74/commits/fffe2d7ee81d90748cdb9867bd55a6eff115af2a) which checks for and adjusts known deprecated ISO 639 codes. After lots of research, it doesn't seem like there's a single method that covers the backward and forwards compatibility that @1ec5 mentions above. Let's let my commit ✅  this comment for now so that this isn't a blocker for getting this plugin released.

langsmith: [This commit](https://github.com/mapbox/mapbox-plugins-android/pull/74/commits/35618ce7a43b0d8960534967246e695636811984) addresses simplified/traditional Chinese selection

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: It can definitely be cleaned up, but here's my`textField` check for now 😕 : https://github.com/mapbox/mapbox-plugins-android/blob/5a1c2854a6935d4ab0ae9a3eb8ede0a65680a7bd/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L36-L42

`{abbr}` check and adjustment: https://github.com/mapbox/mapbox-plugins-android/blob/5a1c2854a6935d4ab0ae9a3eb8ede0a65680a7bd/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L43-L46

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
1ec5: How about [toLanguageTag](https://developer.android.com/reference/java/util/Locale.html#toLanguageTag())?

langsmith: yea, looked into it. For some reason, I think I read that it wouldn't work both ways, but I don't remember exactly why now 😵 . That link you just gave above, _does_ sound like it would work for covering back and forwards compatibility. Will look back into it

langsmith: Using `toLanguageTag()` if api level is above 21.

Both issues brought up in OP have been addressed.

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
langsmith: Now checking for Mapbox-only sources with this commit

https://github.com/mapbox/mapbox-plugins-android/pull/74/commits/0314d08e6d91a87f66aae1adab1981deed6cc740

langsmith:
After lots of refactoring, this is what the test app activity is looking like
ezgif com-resize 2

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Checks whether the map's source is a source provided by Mapbox, rather than a custom source.
+   *
+   * @param singleSource an individual source object from the map
+   * @return true if the source is from a Mapbox vector source, false if it's a
+   * custom Studio data set or from elsewhere.
+   */
+  private boolean sourceIsFromMapbox(Source singleSource) {
+    return singleSource instanceof VectorSource
+      && ((VectorSource) singleSource).getUrl().contains("mapbox://mapbox.mapbox");
```
1ec5: Many Mapbox-designed styles – indeed, many styles designed in Studio – use a single composite source whose URL is `mapbox://mapbox.mapbox-terrain-v2,mapbox.mapbox-streets-v7`. By only checking if the URL begins with `mapbox://mapbox.mapbox`, this code may inadvertently snag sources that are just based on `mapbox.satellite` or `mapbox.terrain`.

[The iOS/macOS map SDK’s implementation](https://github.com/mapbox/mapbox-gl-native/blob/0eb4b5cc0ec494f4b8de6933d2eddf97b12a77f7/platform/darwin/src/MGLVectorSource.mm#L102-L109) considers the source to be the Mapbox Streets source – the only one this plugin can reliably localize – if both of the following conditions hold:

1. The URL scheme is `mapbox`; that is, the URL begins with `mapbox://`.
1. The URL’s hostname is a comma-delimited list containing either `mapbox.mapbox-streets-v7` or `mapbox.mapbox-streets-v6`. (The latter doesn’t matter so much at this point.)

langsmith: Thanks @1ec5 . Will adjust based on your comments. 🚀

langsmith: Although I don't create String arrays like the iOS/macOS map SDK’s implementation, I believe that the way I've handled scheme/hostname check will work just fine: 

https://github.com/mapbox/mapbox-plugins-android/blob/08d5ef3013c52e8e8d4feb57811eeda6b05a4458/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L109-L111

1ec5: Yeah, I think it’s robust enough. It’s unlikely that someone would create an account named `likemapbox` and upload a dataset named `mapbox-streets-v71` to it. It’s also going to be awhile before the cartography team gets a chance to release Mapbox Streets source v60.

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -0,0 +1,193 @@
...
+  public void setMapTextLanguage(String languageToSetMapTo) {
+    for (Source source : mapboxMap.getSources()) {
+      if (sourceIsFromMapbox(source)) {
+        for (Layer layer : mapboxMap.getLayers()) {
+          if (layerHasAdjustableTextField(layer)) {
+            if (((SymbolLayer) layer).getTextField().getValue().contains("name")) {
+              layer.setProperties(textField(String.format("{name_%s}", languageToSetMapTo)));
+            }
+            if (((SymbolLayer) layer).getTextField().getValue().contains("abbr")
+              && !getDeviceLanguage().equals("en")) {
```
1ec5: This assumes the style was set to use `name_en` in the first place, which is true about most Mapbox-designed styles like Mapbox Streets. However, it’d be totally reasonable for a style to use `name` (local language) right out of the gate.

For reference, the iOS and macOS map SDKs preserve the style’s original language (e.g., English) unless the `localizesLabels` feature is enabled. Once the feature is enabled, it [changes every `name` and `name_*` to the device language](https://github.com/mapbox/mapbox-gl-native/blob/e2a4dffcb627498561d66a304ec668f8242b8342/platform/darwin/src/MGLVectorSource.mm#L118-L122), [falling back to `name_en`](https://github.com/mapbox/mapbox-gl-native/blob/e2a4dffcb627498561d66a304ec668f8242b8342/platform/darwin/src/MGLVectorSource.mm#L99) if the device language is unsupported. So whether the style originally used `name_en`, `name`, or `name_zh`, the localization feature will still take effect.

(The Mapbox GL Language plugin for GL JS has a [`defaultLanguage`](https://github.com/mapbox/mapbox-gl-language/blob/9410fd13ce5ef47b90f91bb9af363557614b360b/index.js#L23) option instead of falling back to English.)

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationActivity.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  @OnClick(R.id.localize_fab)
+  public void localizeToggleFab() {
+    if (mapIsLocalized) {
+      localizationPlugin.setMapTextLanguage(ENGLISH_LANGUAGE_CODE);
+      Toast.makeText(this, R.string.map_not_localized, Toast.LENGTH_SHORT).show();
+      mapIsLocalized = false;
+    } else {
+      localizationPlugin.setMapTextLanguage(localizationPlugin.getDeviceLanguage());
```
1ec5: If the device language happens to be a language that the Mapbox Streets source doesn’t support, such as Thai, this plugin will end up changing all the layers to `{name_th}`, producing undesired results. This plugin needs to check that the device language is [supported](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#overview), falling back to either English or the local language. (The iOS and macOS map SDKs fall back to English, whereas the Mapbox GL Language plugin for GL JS falls back to a default language that the developer specifies as an option.)

langsmith: I believe that https://github.com/mapbox/mapbox-plugins-android/pull/74/commits/af59d951965aa060e8d0e91726ac216d92966759  now fixes this. I've added a preferred backup language parameter for the initialization of the plugin.

https://github.com/mapbox/mapbox-plugins-android/blob/af59d951965aa060e8d0e91726ac216d92966759/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L39

I also added a check to make sure that a language is supported by OSM
https://github.com/mapbox/mapbox-plugins-android/blob/af59d951965aa060e8d0e91726ac216d92966759/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L87-L97

If the check detects that the device language isn't supported, then the backup language is checked and set. English is the final fall back if the *backup* language isn't supported by OSM either

https://github.com/mapbox/mapbox-plugins-android/blob/af59d951965aa060e8d0e91726ac216d92966759/plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java#L130-L138

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
langsmith: Removed the ` && !getDeviceLanguage().equals("en")` assumption in this commit https://github.com/mapbox/mapbox-plugins-android/pull/74/commits/86be093d0f65d60d8dd8c3a86287e7ffe2a0109f 

The comment below addresses the checking and fallback functions that I've built in.

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -24,15 +24,21 @@
...
   private MapboxMap mapboxMap;
+  private String selectedBackupLanguage;
 
   /**
    * Create a {@link LocalizationPlugin}.
    *
-   * @param mapboxMap the MapboxMap to apply the localization plugin to
+   * @param mapboxMap      the MapboxMap to apply the localization plugin to
+   * @param backupLanguage the default language for the map to fall back to if the device language is not
+   *                       supported by Open Street Maps.
```
1ec5: Nit: OpenStreetMap supports every language that has an ISO 639 code. It’s the Mapbox Streets source that’s currently limited to [a handful of languages](https://www.mapbox.com/vector-tiles/mapbox-streets-v7/#overview).

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -73,21 +79,36 @@ public String getDeviceLanguage() {
...
       if (sourceIsFromMapbox(source)) {
         for (Layer layer : mapboxMap.getLayers()) {
           if (layerHasAdjustableTextField(layer)) {
-            if (((SymbolLayer) layer).getTextField().getValue().contains("name")) {
-              layer.setProperties(textField(String.format("{name_%s}", languageToSetMapTo)));
-            }
-            if (((SymbolLayer) layer).getTextField().getValue().contains("abbr")
-              && !getDeviceLanguage().equals("en")) {
+            if (((SymbolLayer) layer).getTextField().getValue().contains("name")
+              || ((SymbolLayer) layer).getTextField().getValue().equals("abbr")) {
```
1ec5: If the language is English and the text field is `{abbr}`, we should use `{abbr}`, because that field is written in English.

langsmith:
Per @1ec5 's suggestion, let's add installation instructions to [this help page](https://www.mapbox.com/help/change-language/) before releasing this plugin

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
langsmith: Fixed. Thanks

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -0,0 +1,234 @@
...
+   */
+  private void adjustMapTextLayers(String languageToSetMapTo) {
+    for (Source source : mapboxMap.getSources()) {
+      if (sourceIsFromMapbox(source)) {
+        for (Layer layer : mapboxMap.getLayers()) {
+          if (layerHasAdjustableTextField(layer)) {
+            if (((SymbolLayer) layer).getTextField().getValue().contains("{name")
+              || !getDeviceLanguage().equals("en") && ((SymbolLayer) layer).getTextField().getValue().contains("{abbr}")
+              ) {
+              layer.setProperties(textField(String.format("{name_%s}", languageToSetMapTo)));
```
1ec5: If `text-field` was originally set to something like `{name_en} ({name})`, this would replace it with just `{name_en}` for English speakers. I think it would be better to get the text field value as a string, then find and replace `{name}` and `{name_*}` with `{name_xy}` (where _xy_ is `languageToSetMapTo`), then set the modified string as the text field.

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Checks whether a single map layer has a textField that could potentially be localized to the device's language.
+   *
+   * @param singleLayer an individual layer from the map
+   * @return true if the layer has a textField eligible for translation, false if not.
+   */
+  private boolean layerHasAdjustableTextField(Layer singleLayer) {
+    return singleLayer instanceof SymbolLayer && (((SymbolLayer) singleLayer).getTextField() != null
+      && (((SymbolLayer) singleLayer).getTextField().getValue() != null
+      && !(((SymbolLayer) singleLayer).getTextField().getValue().isEmpty())));
```
1ec5: As tail work, consider making this plugin also affect layers whose `text-field` property is set to a function. In the Mapbox Streets style, for example, U.S. state labels are a function with stops for either `{name_en}` or `{abbr}` based on the zoom level.

Without accounting for functions, there will be some inconsistency in state and country labels at low zoom levels. However, expressions will be available in the next release, so it might be worth spending much effort accommodating functions in the meantime.

/ref mapbox/mapbox-gl-native#10713 mapbox/mapbox-gl-native#10944

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
cammace: I addressed this issue by using a regrex and replacing only the {name_*} part in textfields.

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
cammace: > However, expressions will be available in the next release, so it might be worth spending much effort accommodating functions in the meantime.

I'd prefer merging this PR and fixing this in a different branch in the future when we switch to 6.0.0.

zugaldia: @cammace Is there a ticket in this repo where we're tracking this tail work?

cammace: I ticketed this out in https://github.com/mapbox/mapbox-plugins-android/issues/301

plugins/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/LocalizationPlugin.java
1ec5: It turns out that commit doesn’t actually filter by source. Instead, it reruns the code once for each source, meaning that each layer gets processed one time per source. The Android map SDK apparently lacks a way to determine the source that a layer is using: mapbox/mapbox-gl-native#12691.

PR #98, opened 2017-09-01
Don't use bearing for navigation if it doesn't exist

### LINKED ISSUE #97: location-layer-plugin bearing changes while standing still
### Issue
When standing still (high priority) then all sudden the bearing changes.
Could it be that the last location doesn't have bearing or bad bearing? Or does it need to filter the values a bit more?

Note that this is for the navigation mode and not compass.

### Reproducability
Easily reproduced in a car or public transport with a phone.  Different trackers and playing back it as mock locations is way harder. 

Image of the issue (for the viewer, I'm on a bus and the beginning "left" looking position is the correct way we're travelling and then comes a red light. It should never point upwards). 
videotogif_2017 08 31_18 09 14

Note totally unrelated: definitely looking for apps/plugins for either osx, windows or android to easily create these gifs from video records of the screen.

davols:
Closes #97 

The default value for [`location.getBearing()`](https://developer.android.com/reference/android/location/Location.html#getBearing()) is 0.0 if it does not exist. Which caused the issue, usually when standing still in `NAVIGATION` mode. 
Tiny change


cammace:
Moving forward with this PR, going to close and open a new PR in an attempt to get CI to run.

PR #114, opened 2017-09-28
Cancelable downloads

tobrun:
WIP (still need to fine tune some interactions).This PR adds a cancel button to the notification and allows to cancel ongoing offline downloads. When such a situation occurs we dispatch a broadcast indicating that the download was cancelled.


tobrun:
Since OP, besides basic cancellation through notification, 
now supporting cancellation through bounded activity:

ezgif com-video-to-gif 34

PR #171, opened 2017-11-19
Places plugin

cammace:
As part of the Mapbox Java 3.0 release, we are removing the geocoder widget and replacing it with a new, feature rich, version. The places plugin will eventually (not all in this PR) include a location picker activity, place details bottom sheet view, and an autocomplete activity (this PR).

First, you have two autocomplete modes, one with the cards, and the other full screen. You'll pick this in this intent builder among other options as seen below:

```java
Intent intent = new PlaceAutocomplete.IntentBuilder(PlaceAutocomplete.MODE_CARDS)
      .limit(10)
      .backgroundColor(Color.parseColor("#EEEEEE"))
      .injectPlaces(<List of CarmenFeatures>)
      .build(AutocompleteLauncherActivity.this);
    startActivityForResult(intent, REQUEST_CODE_AUTOCOMPLETE);
```

when the activity starts it looks like this:

ezgif com-crop

The cards which include bright blue text are the CarmenFeatures which were injected using the IntentBuilder, Useful for including users home/work address or quick access to other favorite places. 

When the user selects an item from the list, the activity finishes, and the picked CarmenFeature is received through `onActivityResult`. 

cc: @mapbox/android


langsmith:
Love it @cammace . I'm going to call it `com.mapbox.mapboxsdk:mapbox-android-plugin-places` for now as I set up docs for this, but let me know whatever your final naming decision is.

langsmith:
Love it @cammace . I'm going to call it `com.mapbox.mapboxsdk:mapbox-android-plugin-places` for now as I set up docs for this, but let me know whatever your final naming decision is.

cammace:
Damn your quick @langsmith :) not sure what naming we want to go with here either.

gradle/dependencies.gradle
```diff
@@ -8,27 +8,28 @@ ext {
...
-      timber             : '4.5.1',
-      awsPolly           : '2.3.8',
-      testRunnerVersion  : '1.0.1',
-      espressoVersion    : '3.0.1',
-      spoonRunner        : '1.6.2',
-      commonsIO          : '2.5',
-      robolectric        : '3.4.2',
-      lifecycleCompiler  : '1.0.0',
-      lifecycleRuntime   : '1.0.3'
+      mapboxMapSdk     : '5.2.0-beta.4',
```
Guardiola31337: To avoid conflicts with Lost, we should bump `mapboxMapSdk` version to `5.2.0`.

> Per https://github.com/mapbox/mapbox-gl-native/pull/9488 the location provider is optional (we're not including Lost within the SDK anymore) and in https://github.com/mapbox/mapbox-gl-native/pull/10355 we bumped the Lost version to `3.0.4` https://github.com/mapbox/mapbox-gl-native/blob/ffea8e145f5b67bbeb7aec46c77d73b95ef6ea31/platform/android/dependencies.gradle#L23 which is the one that MAS `2.2.9` uses.

PR #186, opened 2017-11-30
Makes the accuracy ring a circle layer rather than fill layer

### LINKED ISSUE #79: Accuracy circle layer
https://github.com/mapbox/mapbox-gl-native/pull/9426 looks to add support for circle pitch alignment in runtime styling thus unlocking the ability to simplify the logic for the accuracy ring and animate it. 

We are currently using a fill layer since it lays flat even when the maps tilted but with this property unlocked it should become a circle layer. Technically all the layer really could become a circle layer but for the customizability available as far as using drawable for the location icon, I'm thinking they should remain symbol layers.

cc: @tobrun @zugaldia

cammace:
Closes #79 

- [x] Still need to add logic to adjust when the user zooms in and out.


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -178,11 +174,14 @@ private void addNavigationLayer(Drawable navigationDrawable) {
...
 
   private void addAccuracyLayer(float accuracyAlpha, @ColorInt int accuracyColor) {
-    FillLayer locationAccuracyLayer = new FillLayer(
-      LocationLayerConstants.LOCATION_ACCURACY_LAYER, LocationLayerConstants.LOCATION_ACCURACY_SOURCE
+    CircleLayer locationAccuracyLayer = new CircleLayer(
+      LocationLayerConstants.LOCATION_ACCURACY_LAYER, LocationLayerConstants.LOCATION_SOURCE
     ).withProperties(
-      PropertyFactory.fillColor(accuracyColor),
-      PropertyFactory.fillOpacity(accuracyAlpha)
+      PropertyFactory.circleColor(accuracyColor),
```
tobrun: instead of the whole `PropertyFactory.circleColor(accuracyColor)` you can statically import `circleColor(accurancyColor)` instead. Reads a bit more easy.

cammace:
I'm blocked, trying to fix the zoom issue and the circle layer remaining with a radius that's X meters no matter what zoom level. Currently, the circle sizing is based on the viewport instead of the map.

Issue #149, opened 2017-11-04
LocationLayer 0.3.0


tobrun:
### LocationLayer 0.3.0
  - [x] ~~when location layer is created listen to style reloading, recreate the LocationLayer, the locaiton layer has the same lifetime as the associated style~~  -> #196 
  - [x] ~~applyStyle needs to call into setProperties at runtime and not recreate LocationLayer~~ -> #196 
  - [x] ~~Expose api for last known location https://github.com/mapbox/mapbox-plugins-android/issues/136~~  -> #199 
  - [ ] Optimise for cases as #126. 
 - [x] Fix test app issues as https://github.com/mapbox/mapbox-plugins-android/issues/89.
 - [x] ~~expose click events https://github.com/mapbox/mapbox-plugins-android/issues/134~~ -> #198 
 - [x] rework sources to incorporate the full location model into 1 source 
-  [x] ~~add Circle Accuracy Layer https://github.com/mapbox/mapbox-plugins-android/issues/79~~ -> https://github.com/mapbox/mapbox-plugins-android/pull/186
 - [x] ~~keep source/layer in memory instead of calling `mapboxMap#getSource/getLayer`~~-> #196
 - [x]  ~~redo apply style so the plugin calls into `Layer.setProperties`, should fix https://github.com/mapbox/mapbox-plugins-android/issues/87.~~ -> #196
 - [ ] expose api to tweak animation, related to https://github.com/mapbox/mapbox-plugins-android/issues/53
 - [ ] if location isn't found in the camera bounds do not animate. When the it returns, move instantly. (don't animate from previous location) https://github.com/mapbox/mapbox-plugins-android/issues/64. Good use-case for something as a custom source, for now use a different tracking mechanism. 
 - [ ] add stale state https://github.com/mapbox/mapbox-plugins-android/issues/58
 - [ ] add symbol layer text api https://github.com/mapbox/mapbox-plugins-android/issues/59
 - [ ] update design, see design repo
 - [ ] tracking mode changes in https://github.com/mapbox/mapbox-plugins-android/issues/217

@mapbox/navigation-android @mapbox/android @lilykaiser

cammace:
I've moved the remaining issues to the 0.4.0 milestone, we can use the individual tickets to track the issues.

PR #196, opened 2017-12-02
Rework location layer abstraction

tobrun:
Refs #149

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
@@ -2,12 +2,14 @@ package com.mapbox.mapboxsdk.plugins.locationlayer;
 
 import android.content.Context;
 import android.content.res.TypedArray;
+import android.graphics.PointF;
 import android.graphics.drawable.Drawable;
 import android.support.annotation.ColorInt;
 import android.support.annotation.Nullable;
 import android.support.annotation.StyleRes;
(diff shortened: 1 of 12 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -2,199 +2,240 @@
...
+
+      drawableResId = typedArray.getResourceId(R.styleable.LocationLayer_bearingDrawable, -1);
+      styleBearing(ContextCompat.getDrawable(context, drawableResId));
+
+      drawableResId = typedArray.getResourceId(R.styleable.LocationLayer_navigationDrawable, -1);
+      styleNavigation(ContextCompat.getDrawable(context, drawableResId));
+
+      float accuracyAlpha = typedArray.getFloat(R.styleable.LocationLayer_accuracyAlpha, 0.15f);
+      if (accuracyAlpha < 0 || accuracyAlpha > 1) {
+        throw new RuntimeException("Location layer accuracy alpha value must be between 0.0 and 1.0.");
```
cammace: can be more specific here with `IllegalArgumentException`

tobrun: I did not really change anything to that code block, was copied from previous implementation. Seeing that you are already addressing it in https://github.com/mapbox/mapbox-plugins-android/pull/186/files#diff-6b76d34c67041b321a2ce642ab12a240R75, I think we can ignore this for now?

PR #204, opened 2017-12-02
Change default offline download configuration

tobrun:
This PR adapts the default used values in the testapp to be match realistic offline regions.


cammace:
`setMin` isn't available until API level 26 and above. The note on setting the minZoom would be better suited for the Mapbox Demo App.

PR #220, opened 2017-12-21
POC nav mode

tobrun:
This PR contains a small POC of a third person navigation mode, in this mode we have a fixed ImageView over the map and animate the map underneath. 

This PR tries to showcase:
 -  using a focal point with `MapboxMap#getUiSettings()#setFocalPoint(PointF)`. As a result all gesture interactions will occur around this point instead of the focal point of the gesture itself (eg. center point between two fingers while pinch zooming). The calculation isn't completely correct as it wobbles a bit, and we would need to add scaling to it, but you get the picture.

    -  ezgif com-video-to-gif 23

 - you can listen to camera changes and validate if the tilt of the camera has changed. This allows to provide a hook to transform the location puck. Atm the transformation is incorrect but good enough to showcase it's feasibility.

     -  ezgif com-video-to-gif 24

  -  all animations in above gifs were done with Android SDK animators, it contains an example how to orchestrate multiple of them together to create an animation set. Atm animation duration is calculated based on the time it took for the previous location update to be triggerred. There will be better ways of getting a more smooth animation. 

cc @mapbox/navigation-android

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
@@ -39,6 +39,7 @@ public class LocationLayerModesActivity extends AppCompatActivity implements OnM
   private LocationEngine locationEngine;
   private MapboxMap mapboxMap;
   private boolean customStyle;
+  private boolean navigationTrackingState;
 
   @Override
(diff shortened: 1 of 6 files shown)


ericrwolfe:
Awesome! 😍

vladgheorghe:
awesome ! looking forward to the snapshot version

lucianboboc:
Woohoo, let’s go with the merge 😋

tobrun:
fwiw this was a proof of concept, we are looking implementing this functionality in a different manner.

PR #227, opened 2018-01-02
Location Layer Improvements

cammace:


psinetron:
LocationLayerPlugin.java have a 
private Location location variable;
if variable used only on 
public void onCameraMove() method. But this variable every time is null. The accuracy radius is hidden when we move the camera.
Need to update.

psinetron:
Good request. Fix bug with emty location on LocationLayerPlugin.java

cammace:
@psinetron The changes that were in this PR were merged in https://github.com/mapbox/mapbox-plugins-android/pull/228. They are now in master.

PR #245, opened 2018-01-16
Location layer asset update

cammace:
#This PR optimizes/updates some of the location layer assets. The main difference is now we are adding a more 3D effect to the icons so that the user location pops out more and is easier to find (especially on a white background). 

ezgif com-video-to-gif 2

I've also fixed an issue with the accuracy layer getting initialized with the default radius of 5 and then immediately going back to zero since the `location` variable was null. `lastLocation` variable has been removed since `location` is virtually identical.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -29,9 +29,13 @@ import static android.Manifest.permission.ACCESS_FINE_LOCATION;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.ACCURACY_LAYER;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.BEARING_LAYER;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.COMPASS_UPDATE_RATE_MS;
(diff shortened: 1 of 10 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -377,6 +377,14 @@ public void onCompassAccuracyChange(int compassStatus) {
     // Currently don't handle this inside SDK
   }
 
+  private void addCameraListener() {
+    if (locationLayerMode == LocationLayerMode.NAVIGATION) {
+      mapboxMap.removeOnCameraMoveListener(this);
+      return;
```
tobrun: it's a bit strange that a method called #addCameraListener is actually removing the listener in some cases. Either rename the method to something fitting or decouple the removal into a separate codeblock

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -496,6 +504,8 @@ private void setNavigationLayerVisibility(boolean visible) {
   @Override
   public void onCameraMove() {
     locationLayer.updateAccuracyRadius(location);
+    locationLayer.updateForegroundOffset(mapboxMap.getCameraPosition().tilt);
+    locationLayer.updateForegroundBearing((float) mapboxMap.getCameraPosition().bearing);
```
tobrun: reuse the CameraPosition object, now we are calling twice into c++ code while we can get away with just once.

cammace:
Thanks for the review, made the changes @tobrun

PR #262, opened 2018-01-29
Place picker

cammace:
This PR introduces the place picker module to the places plugin. 

ezgif com-video-to-gif


plugin-places/src/main/java/com/mapbox/plugins/places/picker/viewmodel/PlacePickerViewModel.java
```diff
@@ -0,0 +1,47 @@
...
+    if (response.body().features().isEmpty()) {
+      results.setValue(null);
+      return;
+    }
+    results.setValue(response.body().features().get(0));
+  }
+
+  @Override
+  public void onFailure(Call<GeocodingResponse> call, Throwable t) {
+
```
langsmith: Log or do something here? Or comment saying that this was left empty on purpose?

Issue #58, opened 2017-07-17
Location stale state indicator


cammace:
The icon color should fade to gray when the user location hasn't been updated for x amount of time.

PR #264, opened 2018-01-29
Initial stale logic added

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/58

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/StaleStateRunnable.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/StaleStateRunnable.java
@@ -6,23 +6,23 @@ import android.support.annotation.NonNull;
 import java.util.ArrayList;
 import java.util.List;
 
+/**
+ * Class controls the location layer stale state when the {@link android.location.Location} hasn't
+ * been updated in 'x' amount of time. {@link LocationLayerOptions#staleStateDelay()} can be used to
+ * control the amount of time before the locations considered stale.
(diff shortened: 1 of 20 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/StaleStateRunnable.java
```diff
@@ -0,0 +1,56 @@
...
+
+  private static StaleStateRunnable instance;
+
+  private static final int DEFAULT_DELAY_TIME_TILL_STALE = 10000;
+
+  private List<OnLocationStaleListener> onLocationStaleListeners;
+  private Handler handler;
+  private long delayTime = DEFAULT_DELAY_TIME_TILL_STALE;
+
+  static StaleStateRunnable getInstance() {
```
tobrun: is there are reason why this class is a singleton? can't we get away with using a final field in the plugin instead?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/StaleStateRunnable.java
```diff
@@ -0,0 +1,56 @@
...
+    if (instance == null) {
+      instance = new StaleStateRunnable();
+    }
+    return instance;
+  }
+
+  private StaleStateRunnable() {
+    onLocationStaleListeners = new ArrayList<>();
+    handler = new Handler();
```
tobrun: these can be directly initialised as part of the field declaration + they can be made final

Issue #183, opened 2017-11-28
Enable Customisation for Location Layer Plugin


nikiizvorski:
```
@RequiresPermission(anyOf = {Manifest.permission.ACCESS_FINE_LOCATION, Manifest.permission.ACCESS_COARSE_LOCATION})
  public void setLocationLayerEnabled(@LocationLayerMode.Mode int locationLayerMode) {
    if (locationLayerMode != LocationLayerMode.NONE) {
      locationLayer.setLayerVisibility(true);

      // Set an initial location if one is available and the locationEngines not null
      if (locationEngine != null) {
        setLastLocation();
        locationEngine.addLocationEngineListener(this);
      }

      if (locationLayerMode == LocationLayerMode.COMPASS) {
        setLinearAnimation(false);
        setNavigationEnabled(false);
        setMyBearingEnabled(true);
      } else if (locationLayerMode == LocationLayerMode.NAVIGATION) {
        setMyBearingEnabled(false);
        setNavigationEnabled(true);
      } else if (locationLayerMode == LocationLayerMode.TRACKING) {
        setLinearAnimation(false);
        setMyBearingEnabled(false);
        setNavigationEnabled(false);
      }
    } else {
      // Check that the mode isn't already none
      if (this.locationLayerMode != LocationLayerMode.NONE) {
        disableLocationLayerPlugin();
      }
    }
    this.locationLayerMode = locationLayerMode;
  }
```

Please enable any customisation on these methods since you depricated everything should be the time to update everything.... There is currently no good way to update user location icon.

Solution: some methods that we can do that and enable the structure of the code

Solution2: Requested seems like there is something wrong with that plugin and there is no location tracking what so ever. On never devices and old one is depricated

cammace:
You can apply new styles to your location layer icons using a custom style and using the `LocationLayerPlugin.applyStyle`. The custom style would look like so:

```
<style name="CustomLocationLayer" parent="LocationLayer">
    <item name="backgroundDrawable">@drawable/custom_user_stroke_icon</item>
    <item name="foregroundDrawable">@drawable/custom_user_icon</item>

    <item name="bearingDrawable">@drawable/custom_user_arrow</item>
    <item name="navigationDrawable">@drawable/custom_user_puck_icon</item>

    <item name="accuracyAlpha">0.15</item>
    <item name="accuracyColor">#FF82C6</item>
  </style>
```

[...]

PR #267, opened 2018-01-31
adds location layer options class

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/183

Allows you to set style dynamically using the `LocationLayerOptions` class. Also adds all the necessary attributes that were settable in the now deprecated `MyLocationViewSettings`.

### DIFF
--- a/app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerTest.java
+++ b/app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerTest.java
@@ -1,5 +1,6 @@
 package com.mapbox.mapboxsdk.plugins.locationlayer;
 
+import android.content.Context;
 import android.location.Location;
 import android.support.test.espresso.Espresso;
 import android.support.test.espresso.IdlingResourceTimeoutException;
@@ -47,10 +48,9 @@ public class LocationLayerTest {
   public ActivityTestRule<LocationLayerModesActivity> rule = new ActivityTestRule<>(LocationLayerModesActivity.class);
 
   private OnMapReadyIdlingResource idlingResource;
-  private MapboxMap mapboxMap;
   private LocationLayerPlugin locationLayerPlugin;
-
-  Location location;
+  private MapboxMap mapboxMap;
+  private Location location;
 
   @Before
   public void beforeTest() {
@@ -83,7 +83,7 @@ public class LocationLayerTest {
     executeLocationLayerTest(new LocationLayerPluginAction.onPerformLocationLayerAction() {
       @Override
       public void onLocationLayerAction(LocationLayerPlugin locationLayerPlugin, MapboxMap mapboxMap,
-                                        UiController uiController) {
+                                        UiController uiController, Context context) {
         locationLayerPlugin.setLocationLayerEnabled(LocationLayerMode.TRACKING);
         assertTrue(mapboxMap.getSource(LOCATION_SOURCE) != null);
       }
@@ -95,7 +95,7 @@ public class LocationLayerTest {
     executeLocationLayerTest(new LocationLayerPluginAction.onPerformLocationLayerAction() {
       @Override
       public void onLocationLayerAction(LocationLayerPlugin locationLayerPlugin, MapboxMap mapboxMap,
-                                        UiController uiController) {
+                                        UiController uiController, Context context) {
         locationLayerPlugin.setLocationLayerEnabled(LocationLayerMode.TRACKING);
         assertTrue(mapboxMap.getLayer(ACCURACY_LAYER) != null);
         assertTrue(mapboxMap.getLayer(BACKGROUND_LAYER) != null);
@@ -109,7 +109,7 @@ public class LocationLayerTest {
     executeLocationLayerTest(new LocationLayerPluginAction.onPerformLocationLayerAction() {
       @Override
       public void onLocationLayerAction(LocationLayerPlugin locationLayerPlugin, MapboxMap mapboxMap,
-                                        UiController uiController) {
+                                        UiController uiController, Context context) {
         locationLayerPlugin.setLocationLayerEnabled(LocationLayerMode.COMPASS);
         assertTrue(mapboxMap.getLayer(ACCURACY_LAYER) != null);
         assertTrue(mapboxMap.getLayer(BACKGROUND_LAYER) != null);
@@ -124,7 +124,7 @@ public class LocationLayerTest {
     executeLocationLayerTest(new LocationLayerPluginAction.onPerformLocationLayerAction() {
       @Override
       public void onLocationLayerAction(LocationLayerPlugin locationLayerPlugin, MapboxMap mapboxMap,
-                                        UiController uiController) {
+                                        UiController uiController, Context context) {
         locationLayerPlugin.setLocationLayerEnabled(LocationLayerMode.COMPASS);
         assertTrue(mapboxMap.getLayer(NAVIGATION_LAYER) != null);
       }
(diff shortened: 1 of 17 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/StaleStateRunnable.java
```diff
@@ -27,10 +29,18 @@ private StaleStateRunnable() {
...
+  void addOnLocationStaleListener(@NonNull OnLocationStaleListener onLocationStaleListener) {
     onLocationStaleListeners.add(onLocationStaleListener);
   }
 
+  void removeOnLocationStaleListener(@Nullable OnLocationStaleListener onLocationStaleListener) {
+    if (onLocationStaleListener == null) {
+      onLocationStaleListeners.clear();
+      return;
+    }
+    onLocationStaleListeners.remove(onLocationStaleListener);
```
tobrun: this method doesn't do what is expected from its name when a null is passed.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -431,11 +439,13 @@ private void disableLocationLayerPlugin() {
    * position.
    */
   @SuppressWarnings( {"MissingPermission"})
-  private void setLastLocation() {
+  private void setLastLocation(int locationLayerMode) {
```
tobrun: why does the end developer need to pass in the mode? isn't this part of the state managed by the plugin itself?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -260,6 +266,7 @@ public LocationEngine getLocationEngine() {
    */
   @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
   public void onStop() {
+    StaleStateRunnable.getInstance().removeOnLocationStaleListener(null);
     StaleStateRunnable.getInstance().onStop();
```
tobrun: Calling getInstance twice while one time is sufficient, either assign result/reuse or refactor StateStateRunnable to encapsulate behaviour into one method. If `removeOnLocationStaleListener ` is only used in this code path the latter suggestion would make most sense.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
```diff
@@ -0,0 +1,209 @@
...
+import android.support.annotation.DrawableRes;
+import android.support.annotation.FloatRange;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.StyleRes;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class LocationLayerOptions implements Parcelable {
```
tobrun: missing javadoc on public API from this class.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/Utils.java
```diff
@@ -37,20 +44,40 @@ static Bitmap getBitmapFromDrawable(Drawable drawable) {
...
+      Bitmap bitmap = Bitmap.createBitmap(drawable.getIntrinsicWidth(),
+        drawable.getIntrinsicHeight(), Bitmap.Config.ARGB_8888);
       Canvas canvas = new Canvas(bitmap);
       drawable.setBounds(0, 0, canvas.getWidth(), canvas.getHeight());
       drawable.draw(canvas);
       return bitmap;
     }
   }
 
+  static Bitmap setShadow(Drawable drawable, float elevation) {
```
tobrun: setShadow is not a setter

app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerTest.java
```diff
@@ -153,6 +164,38 @@ public void onLocationLayerAction(LocationLayerPlugin locationLayerPlugin, Mapbo
...
+
+  @Test
+  public void whenDrawableChanged_continuesUsingStaleIcons() throws Exception {
+
+  }
+
+  @Test
+  public void whenModeChanged_remainsUsingStaleIcons() throws Exception {
+
+  }
```
tobrun: empty tests

app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerTest.java
```diff
@@ -44,6 +50,8 @@
   private MapboxMap mapboxMap;
   private LocationLayerPlugin locationLayerPlugin;
 
+  Location location;
```
tobrun: make private?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
cammace: This method isn't exposed to the end dev, it's purpose is to attempt at getting a location value immediately when a mode is set.

cammace:
Thanks for the notes @tobrun, I've fixed up the issues and am ready for a second pass. I also forgot to mention that this PR branched from #264, it might be easier if I merge this into that branch for the final review. Let me know what works for you.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
tobrun: That intent is not clear from the method name or the parameter, I would advice coming up with a better name for those.

tobrun: Thanks for clarifying, I'm still a bit confused why you aren't moving this [line](https://github.com/mapbox/mapbox-plugins-android/blob/ae4dc1b3c7f6fba46871ab88f3a5340065c19b95/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java#L179) to the top of the method and just checking against `this.locationLayerMode` instead. Would this break any business logic?

cammace: Good point, initially testing and not seeing any issues with this approach.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -152,12 +152,13 @@ private void initialize() {
    */
   @RequiresPermission(anyOf = {ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION})
   public void setLocationLayerEnabled(@LocationLayerMode.Mode int locationLayerMode) {
+    this.locationLayerMode = locationLayerMode;
```
tobrun: this change allows to also remove the param in `toggleCameraListener`?

PR #292, opened 2018-02-19
Changing LocationEngine

mayardb:
When replacing the Location from LocationLayerPlugin, the LocationLayerPlugin was still listening the previous LocationEngine. I also remove first, then set the new one in `setLocationEngine` method.


tobrun:
@mayardb thank you for this contribution!

cammace:
I moved this PR over to https://github.com/mapbox/mapbox-plugins-android/pull/298 so that CI would run. This PR looks good to me though. As @tobrun mentioned, thanks for contributing.

PR #293, opened 2018-02-19
Map + Location Animators 

danesfeder:
This PR will aim to add new, more powerful APIs for animating the `MapboxMap` camera.  This functionality will include the ability to account for padding, zoom, and tilt when animating towards a targeted `LatLng`.  Can be integrated with the `LocationLayer` to add the same camera tracking functionality found in the currently `MyLocationView` in the Maps SDK. 

cc @tobrun @LukasPaczos


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/camera/MapAnimator.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+
+
+    public MapAnimator build() {
+      return new MapAnimator(animators);
+    }
+  }
+
+  private static float normalizeBearing(float currentBearing, float targetBearing) {
```
cammace: We should just extract this method into the Utils class since I already use the same logic in `LocationLayerPlugin`

danesfeder: @cammace Sure thing.  Just porting this code from https://github.com/mapbox/mapbox-navigation-android/pull/679 - will look to clean up and remove duplicate code as the PR progresses

cammace: awesome

PR #294, opened 2018-02-20
Add Camera and Tracking Modes

danesfeder:


app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
```diff
@@ -93,6 +96,7 @@ public void onMapReady(MapboxMap mapboxMap) {
     locationEngine.activate();
     locationLayerPlugin = new LocationLayerPlugin(mapView, mapboxMap, locationEngine);
     locationLayerPlugin.setOnLocationClickListener(this);
+    locationLayerPlugin.setLocationLayerEnabled(true);
```
tobrun: can this be the default behaviour, so no additional setter need to be called after using the constructor?

tobrun:
the `setLocationLayerStyle` also resembles a lot with `applyStyle`, we need to streamline those api's

danesfeder:
@tobrun We still need to disable linear animation after using navigation mode and going back to a different mode

PR #295, opened 2018-02-20
Localization

cammace:
branched from `ls-localization-plugin` and took a slightly different approach to map localization. `MapLocale` follows the `Locale` logic but adds additional information (mapLanguage and bbox). Still, need to write more test before final merge but would like to hear thoughts on this approach.


plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/MapLocale.java
```diff
@@ -0,0 +1,145 @@
...
+    LOCALE_SET.put(Locale.CHINA, MapLocale.CHINA);
+    LOCALE_SET.put(Locale.TAIWAN, MapLocale.TAIWAN);
+    LOCALE_SET.put(Locale.PRC, MapLocale.PRC);
+    LOCALE_SET.put(Locale.ITALY, MapLocale.ITALY);
+    LOCALE_SET.put(Locale.UK, MapLocale.UK);
+    LOCALE_SET.put(Locale.JAPAN, MapLocale.JAPAN);
+    LOCALE_SET.put(Locale.KOREA, MapLocale.KOREA);
+    LOCALE_SET.put(Locale.GERMANY, MapLocale.GERMANY);
+    LOCALE_SET.put(Locale.FRANCE, MapLocale.FRANCE);
+  }
```
tobrun: I would favor using double brace initalisation above a static block:
```java
private static final Map<Object, Object> MAP = new HashMap<Object, Object>(){{
    put(new Object(), new Object());
    put(new Object(), new Object());
  }};
```

plugin-localization/src/test/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPluginTest.java
```diff
@@ -0,0 +1,62 @@
...
+//  }
+//
+//  @Test
+//  public void getMapLocale_doesReturnCustomLocale() throws Exception {
+//    Locale locale = new Locale("foo", "bar");
+//    MapLocale mapLocale = new MapLocale("name_foo");
+//    LocalizationPlugin.addMapLocale(locale, mapLocale);
+//    MapLocale finalLocale = LocalizationPlugin.getMapLocale(locale);
+//    assertThat(finalLocale.getMapLanguage(), equalTo("name_foo"));
+//  }
```
tobrun: commented code

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/LocalizationPlugin.java
```diff
@@ -1,140 +1,94 @@
...
-        return "zh-Hans";
-      } else if (Locale.getDefault().getLanguage().contains("zh-Hant")
-        || Locale.getDefault().getDisplayName().contains("繁體")) {
-        return "zh";
-      } else {
-        return Locale.getDefault().getLanguage();
-      }
-    }
+
+  public void setMapLanguage() {
```
tobrun: without looking at the implementation it's not clear from the method name / parameters that this sets the default locale.

plugin-localization/src/main/java/com.mapbox.mapboxsdk.plugins.localization/MapLocale.java
cammace: I get this lint warning when moving them into a double brace:

screen shot 2018-02-21 at 10 16 12 am

PR #298, opened 2018-02-21
Changing LocationEngine (#292)

cammace:
When replacing the Location from LocationLayerPlugin, the LocationLayerPlugin was still listening the previous LocationEngine. I also remove first, then set the new one in `setLocationEngine` method.


LukasPaczos:
This is referenced by https://github.com/mapbox/mapbox-plugins-android/pull/307 and will land as a part of https://github.com/mapbox/mapbox-plugins-android/pull/293.

PR #309, opened 2018-02-23
LocationLayerPlugin Javadoc

danesfeder:


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -140,119 +137,138 @@ public void setLocationLayerEnabled(boolean isEnabled) {
...
   public int getCameraMode() {
     return locationLayerCamera.getCameraMode();
   }
 
+  /**
+   * Sets the render mode, which determines how the location updates will be rendered on the map.
+   * <p>
+   * <ul>
+   * <li>{@link RenderMode#NORMAL}: Basic tracking is enabled, bearing ignored</li>
+   * <li>{@link RenderMode#COMPASS}: Tracking the user location with bearing considered from compass</li>
```
LukasPaczos: Can we change `tracking` keyword from `setRenderMode()` doc to just say something along the lines of "show current user location", not to confuse it with the camera tracking? It should be implicit that it'll always be the most recent location sample available.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -355,93 +340,117 @@ public void addOnLocationClickListener(@NonNull OnLocationLayerClickListener loc
...
+   * @since 0.3.0
+   */
   public void removeOnLocationClickListener(@NonNull OnLocationLayerClickListener locationClickListener) {
     onLocationLayerClickListeners.remove(locationClickListener);
   }
 
+  /**
+   * Adds the passed listener that gets invoked when user updates have stopped long enough for the last update
+   * to be considered stale.
+   *
```
LukasPaczos: What about adding a reference to stale timeout setter/option?

PR #313, opened 2018-02-26
Add max / min zoom and padding APIs 

danesfeder:
Still need to add the improved `stop`s to the puck scaling

Also this is currently branched from `update-javadoc`, so includes all the javadoc changes until they are merged.  Still need to address @LukasPaczos comments in https://github.com/mapbox/mapbox-plugins-android/pull/309


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -255,6 +257,27 @@ public void setPadding(int left, int top, int right, int bottom) {
     mapboxMap.setPadding(left, top, right, bottom);
   }
 
+  /**
+   * Sets the maximum zoom level the map can be displayed at.
+   * <p>
+   * The default maximum zoom level is 22. The upper bound for this value is 25.5.
```
LukasPaczos: Can we add a higher default zoom level to prevent jiggling when zoomed in extremely close?

PR #315, opened 2018-02-26
Update CHANGELOG.md

mdakram:
its location layer plugin not localization layer plugin


langsmith:
🤦‍♂️ , thanks @mdakram

PR #327, opened 2018-02-28
Gestures logic for camera tracking, new telemetry library

LukasPaczos:
This PR adds gestures logic on top of camera tracking alongside with the fixes of breaking changes introduced with new telemetry library within the Maps SDK.

Camera mode will be reverted back to `CameraMode.None` whenever movement threshold (of single or multi pointers is exceeded). Both of those values are settable through `LocationLayerOptions`.

Additionally, whenever the camera is in any of the tracking modes the focal point of all gestures will be located close to the user location on the map.

Currently in use is `SNAPSHOT` of the Maps SDK that includes new gestures implementation. We will need to change it to the `6.0.0-beta.3` before `0.5.0-beta` release of the plugin.

/cc @cammace @Guardiola31337


danesfeder:
location-gestures

Issue #110, opened 2017-09-27
Selection dialog


tobrun:
A nice to have feature for the offline plugin is having a ready made solution that makes it easy for end developers to provide a selection dialog to their end users.

![device-2016-12-05-233546](https://cloud.githubusercontent.com/assets/2151639/20906098/c5ce4b36-bb46-11e6-90cb-f427bcf08dd5.png)

PR #329, opened 2018-02-28
Add Offline region selector UI

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/110


cammace:
The initial work for this UI components finished. @tobrun can you give this a look over when you get a chance?

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/offline/OfflineDownloadActivity.java
```diff
@@ -32,174 +19,34 @@
...
-
-  @BindView(R.id.spinner_style_url)
-  Spinner styleUrlView;
-
-  @BindView(R.id.seekbar_min_zoom)
-  SeekBar minZoomView;
-
-  @BindView(R.id.seekbar_max_zoom)
-  SeekBar maxZoomView;
+  private static final int REQUEST_CODE = 9384;
```
tobrun: Can we still keep the form around? This is useful for reproducing client bugs with specific bounds/style. The UI selector could be a separate activity in the plugin overview.

plugin-offline/src/main/AndroidManifest.xml
```diff
@@ -1 +1,10 @@
-<manifest package="com.mapbox.androidsdk.plugins.offline"/>
\ No newline at end of file
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.mapbox.mapboxsdk.plugins.offline">
+
+  <application>
+    <activity
+      android:name="com.mapbox.mapboxsdk.plugins.offline.ui.OfflineActivity"
+      android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|uiMode|screenSize|smallestScreenSize"/>
```
tobrun: have you thought about exposing implicit intents?

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+    "place_label", "state_label", "country_label"
+  };
+
+  private RegionSelectedCallback selectedCallback;
+  private TextView regionNameTextView;
+  private MapboxMap mapboxMap;
+  private RectF boundingBox;
+  private MapView mapView;
+  private View rootView;
+  String regionName;
```
tobrun: missing modifier

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+  private static final String[] SOURCE_LAYER_IDS = new String[] {
+    "place_label", "state_label", "country_label"
+  };
+
+  private RegionSelectedCallback selectedCallback;
+  private TextView regionNameTextView;
+  private MapboxMap mapboxMap;
+  private RectF boundingBox;
+  private MapView mapView;
+  private View rootView;
```
tobrun: one time rootView is accessed, for other occurrences getView() is used. 
This can be done in the same manner for all occurrences.

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/OfflineActivity.java
```diff
@@ -0,0 +1,61 @@
...
+    // Add autocomplete fragment if this is first creation
+    if (savedInstanceState == null) {
+
+      RegionSelectionFragment fragment;
+
+      fragment = RegionSelectionFragment.newInstance();
+
+      getSupportFragmentManager().beginTransaction()
+        .add(R.id.fragment_container, fragment, RegionSelectionFragment.TAG).commit();
+
```
tobrun: TAG is only used once, we can also get away with RegionSelectionFragment.class.getSimpleName() instead

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+    mapView.onCreate(savedInstanceState);
+    mapView.getMapAsync(this);
+    bindClickListeners();
+  }
+
+  public RegionSelectedCallback getSelectedCallback() {
+    return selectedCallback;
+  }
+
+  public void setSelectedCallback(@NonNull RegionSelectedCallback selectedCallback) {
```
tobrun: onAttach is exposed on Fragment for setting callbacks, see https://developer.android.com/training/basics/fragments/communicating.html, no need for specific setter

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+    this.selectedCallback = selectedCallback;
+  }
+
+  @Override
+  public void onMapReady(MapboxMap mapboxMap) {
+    this.mapboxMap = mapboxMap;
+    mapboxMap.addOnCameraIdleListener(this);
```
tobrun: we are never removing this listener

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  @Override
+  public void onDestroyView() {
+    super.onDestroyView();
+    mapView.onDestroy();
+  }
+
+  private RectF getSelectionRegion() {
+    View selectionBoxView = rootView.findViewById(R.id.mapbox_offline_scrim_view);
```
tobrun: since getSelectionRegion is called often, we could keep this view in memory

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
```diff
@@ -0,0 +1,186 @@
...
+      Timber.v("Rendered features empty, attempting to query vector source.");
+      VectorSource source = mapboxMap.getSourceAs("composite");
+      if (source != null) {
+        featureList = source.querySourceFeatures(SOURCE_LAYER_IDS, null);
+      }
+    }
+    if (!featureList.isEmpty() && featureList.get(0).getProperties().has("name")) {
+      return featureList.get(0).getStringProperty("name");
+    }
+    return getString(R.string.mapbox_offline_default_region_name);
```
tobrun: 👍 for determining the region name automatically

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/Constants.java
```diff
@@ -1,6 +1,6 @@
 package com.mapbox.mapboxsdk.plugins.offline;
 
-class Constants {
+public class Constants {
```
tobrun: if possible keep this package private

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/MapboxOffline.java
```diff
@@ -0,0 +1,39 @@
...
+import android.app.Activity;
+import android.content.Intent;
+
+import com.mapbox.mapboxsdk.offline.OfflineTilePyramidRegionDefinition;
+import com.mapbox.mapboxsdk.plugins.offline.ui.OfflineActivity;
+
+import static com.mapbox.mapboxsdk.plugins.offline.Constants.RETURNING_DEFINITION;
+import static com.mapbox.mapboxsdk.plugins.offline.Constants.RETURNING_REGION_NAME;
+
+public class MapboxOffline {
```
tobrun: the name implicates a general entry point to offline/plugin but the class is a utility for all things related to intents. MapboxOfflineIntent would be a better name?

plugin-offline/src/main/res/values/mapbox_strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+  <string name="mapbox_offline_default_region_name">Map </string>
+  <string name="mapbox_plugins_offline_toolbar_primary_text">Select a region</string>
+  <string name="mapbox_plugins_offline_toolbar_secondary_text">Pan and zoom to adjust</string>
+
+  <string name="mapbox_plugins_cd_exit_offline">Exit place picker</string>
```
tobrun: More consistency with resource naming, some use mapbox_offline while others use mapbox_plugins_offline as prefix

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
cammace: I did this because the compiler was warning me `Access to private member of class 'RegionSelectionFragment' requires synthetic accessor call`. Should I not worry about this?

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/OfflineActivity.java
cammace: Public tags provided since devs can directly implement the fragment into their app. We do the same thing in the Places Plugin.

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
cammace: Fragment can be directly used by users rather than using the activity if they need to.

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
cammace: Fixed

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/MapboxOffline.java
cammace: Renamed to `OfflineRegionSelector` to keep consistency between places and this plugin.

plugin-offline/src/main/res/values/mapbox_strings.xml
cammace: Fixed

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/Constants.java
cammace: we only have two public constants in here which can be moved into a separate constant file but I don't see the urgency. Places plugin exposes some constants also.

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/ui/RegionSelectionFragment.java
cammace: It should only be called once if boundingbox is null and when we create the region definition. For some reason when placed in `onViewCreated` the view is still measured at 0.

plugin-offline/src/main/AndroidManifest.xml
cammace: shouldn't that be up to the end developer?

PR #334, opened 2018-03-01
added release automation

cammace:
Automate the release process to speed things up a bit :)


cammace:
@tobrun would love to get this in before our next release so I can sure everything works properly.

PR #340, opened 2018-03-03
Fixed jacoco test reporting

cammace:


plugin-building/build.gradle
```diff
@@ -17,6 +17,14 @@ android {
...
     abortOnError false
   }
+
+  testOptions {
+    unitTests.all {
+      jacoco {
+        includeNoLocationClasses = true
+      }
+    }
+  }
```
tobrun: is there a way we can declare the configuration once and use it for all plugins?

cammace: Previously I had it directly in the `jacoco.gradle` file but since it requires to be inside `android` I found it only really works if it's in each modules build.gradle.

gradle/jacoco.gradle
```diff
@@ -1,45 +1,36 @@
 apply plugin: 'jacoco'
 
 jacoco {
-  toolVersion = pluginVersion.jacoco
+  toolVersion "0.7.4.201502262128"
```
cammace: Hardcoded the version here for now since sonarqube seems to have a hard time parsing the latest versions `.ec` file.

PR #349, opened 2018-03-07
Animator Updates

danesfeder:
Chop for the compass animation updates was coming from how we were choosing the previous target.  
    - I opted to cache the previous values for `bearing` and `latlng` in the `LocationLayerAnimator` rather than the `LocationLayerPlugin`.  To me, this just seemed easier to keep track of. 

Now fixed:
ezgif com-video-to-gif

Right now, the animator between camera modes is instant and does not interpolate.  I see this as something that we should consider fixing because we allow devs to change this setting on-the-fly and it ends up being a jarring UX to have the map jump like this: 

ezgif com-video-to-gif

In order to smooth the transitions though, we will need the current map camera position to use as the "previous" bearing or `LatLng`.  I also think this may mean we need separate animators for the camera / location layer?  Adding animators would be more code, but we'd still be able to get the same effect (one animator driving both the layer / camera) if we ensure they both have the same duration and are played together with `AnimatorSet`

Still TODO:
- [ ] Optimize time of animations / interpolators  
- [ ] Find a way to not cancel transition animations when switching between camera modes 
- [x] Reset camera when camera mode changes
- [x] GPS North camera mode


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -26,41 +30,56 @@ void removeListener(OnAnimationsValuesChangeListener listener) {
...
+//      locationChangeAnimator.setInterpolator(new LinearInterpolator());
+//    } else {
+//      locationChangeAnimator.setInterpolator(new AccelerateDecelerateInterpolator());
+//    }
 
     cancelLocationAnimations();
-    latLngAnimator = new LatLngAnimator(previousLatLng, newLatLng, 1000);
-    gpsBearingAnimator = new BearingAnimator(previousBearing, newLocation.getBearing(), 1000);
-    // FIXME: 22/02/2018 evaluate duration of animation better
+    latLngAnimator = new LatLngAnimator(previousLatLng, targetLatLng, 1000);
```
tobrun: can we replace the constant 1000 with the [following](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/maps/widgets/MyLocationView.java#L998-L1012) logic for now. This should make the animation a bit more continuous.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -26,41 +30,56 @@ void removeListener(OnAnimationsValuesChangeListener listener) {
...
+      previousCompassBearing = targetCompassBearing;
+    }
+
+    float previousBearing = getPreviousCompassBearing();
+
     cancelCompassAnimations();
-    compassBearingAnimator = new BearingAnimator(previousCompassBearing, targetCompassBearing, 1000);
-    // FIXME: 22/02/2018 evaluate duration of animation better
+    compassBearingAnimator = new BearingAnimator(previousBearing, targetCompassBearing, 1000);
+    compassBearingAnimator.setInterpolator(new LinearInterpolator());
```
tobrun: no need to set LinearInterpolator specifically, this is the default.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -106,12 +125,49 @@ void cancelAllAnimations() {
...
+      previousLatLng = (LatLng) latLngAnimator.getAnimatedValue();
+    } else {
+      previousLatLng = new LatLng(previousLocation);
+    }
+    return previousLatLng;
+  }
+
+  private float getPreviousGpsBearing() {
+    float previousBearing;
+    if (gpsBearingAnimator != null) {
```
tobrun: the animator is never set to null? those this need to happen as part of cancel?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -106,12 +125,49 @@ void cancelAllAnimations() {
...
+      previousBearing = (float) gpsBearingAnimator.getAnimatedValue();
+    } else {
+      previousBearing = previousLocation.getBearing();
+    }
+    return previousBearing;
+  }
+
+  private float getPreviousCompassBearing() {
+    float previousBearing;
+    if (compassBearingAnimator != null) {
```
tobrun: ditto

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -26,41 +30,56 @@ void removeListener(OnAnimationsValuesChangeListener listener) {
...
+    // TODO create animator duration / interpolator
+//    float speed = location == null ? 0 : location.getSpeed();
+//
+//    locationChangeAnimator.setDuration(linearAnimation || speed > 0
+//      ? getLocationUpdateDuration() : LocationLayerConstants.LOCATION_UPDATE_DELAY_MS);
+//    if (linearAnimation || speed > 0) {
+//      locationChangeAnimator.setInterpolator(new LinearInterpolator());
+//    } else {
+//      locationChangeAnimator.setInterpolator(new AccelerateDecelerateInterpolator());
+//    }
```
tobrun: remove commented code?

tobrun:
@danesfeder 

> Right now, the animator between camera modes is instant and does not interpolate. I see this as something that we should consider fixing because we allow devs to change this setting on-the-fly and it ends up being a jarring UX to have the map jump like this:

The issue is not the interpolator, it's the set duration.  When animating from one location to another using linear is the most logical one and should be default imo. The only cases where another interpolator could be used to finetune the experience would be with when speed from previous update = 0 and new speed > 0 (AccelerateInterpolator) or when speed from new update = 0 and previous speed > 0 (DecelerateInterpolator)

cc @LukasPaczos

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
danesfeder: Ah nice 👍

danesfeder:
@cammace @tobrun @LukasPaczos 

Added a method to reset the camera animators when switching between camera modes (interpolates between current camera position and current target).  Also added the gps north work:

ezgif com-video-to-gif 1

Only two remaining tasks now:  
1.  The duration of the animation is still set to `1000 millis` -- something weird is going on with the location updates.  I'm seeing 2 - 3 updates from the listener for each mock location from pathfinder.  This looks like an issue originating from the new location library in maps `6.0`

2.  When we transition between camera modes, the animation transitioning to the new camera mode is cancelled and interrupted until finally finishes (you can see this in the gif above).  We need to come up with some sort of system to basically say "okay, we are done transitioning now, you can cancel")

LukasPaczos:
@danesfeder great work so far!
Just wanted to add, that I feel like `2.` will be much less of an issue when we resolve `1.`

PR #365, opened 2018-03-14
Offline downloads happen on a separate thread now.

cammace:
This Introduces an OfflineDownloadThread class which handles the offline region downloading and limits the number of callbacks that occur on the main thread (mainly status updates). I'd like to take this a step further in a different branch by allowing multiple regions to be downloaded at the same time.


cammace:
This plugin is now blocked by https://github.com/mapbox/mapbox-gl-native/issues/11461. While the issue is being resolved, I'll be continuing work on cleaning this PR up and documenting the plugin in preparation for release.

cammace:
I'm ready for another review on this PR when you get a chance @tobrun

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/offline/OfflineRegionDetailActivity.java
```diff
@@ -188,36 +188,36 @@ public void onFabClick(View view) {
   }
 
   @Override
-  public void onCreate(OfflineDownloadOptions offlineDownload) {
-    Timber.e("OfflineDownloadOptions created %s", offlineDownload.hashCode());
+  public void onCreate(DownloadOptions offlineDownload) {
```
langsmith: Maybe change the parameter name to `downloadOptions` or something without the word offline in it if the class name is no longer going to have the word `Offline` in it...

plugin-offline/src/main/AndroidManifest.xml
```diff
@@ -6,5 +6,7 @@
     <activity
       android:name="com.mapbox.mapboxsdk.plugins.offline.ui.OfflineActivity"
       android:configChanges="keyboard|keyboardHidden|orientation|screenLayout|uiMode|screenSize|smallestScreenSize"/>
+
```
langsmith: nit line

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/OfflineRegionSelector.java
```diff
@@ -15,6 +17,32 @@ private OfflineRegionSelector() {
     // No Instances
   }
 
+  public static DownloadOptions getDownloadOptions(final Intent data, byte[] metadata) {
```
langsmith: Add a ticket to this repo documenting that explanatory javadocs should be written for this class and its methods?

tobrun:
A couple of more things that needs to be fixed before we can merge this PR:

##### Cancel 

Clicking cancel does stop the download but doesn't dismiss the notification

ezgif com-video-to-gif 56

##### MapSnapshot

I create two downloads with different styles but the same snapshot is used

ezgif com-gif-maker 2

##### Detail activity

The boundable detail activity is broken

ezgif com-video-to-gif 57

##### Multiple threads

Since there still some more work to be done before we can merge this let's move to a single thread model instead of creating a new thread for each download.

plugin-offline/src/main/java/com/mapbox/mapboxsdk/plugins/offline/OfflineRegionSelector.java
cammace: https://github.com/mapbox/mapbox-plugins-android/issues/385 addresses this.

cammace:
Closing infavor of #407

Issue #369, opened 2018-03-16
Location Layer Proguard Rules


SamiPatanen:
When I try to build an APK with proguard enabled, the build fails with the following errors:
```
Warning: com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerOptions: can't find referenced class com.google.auto.value.AutoValue
Warning: com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerOptions$Builder: can't find referenced class com.google.auto.value.AutoValue$Builder
Warning: com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerOptions$Builder: can't find referenced class com.google.auto.value.AutoValue
Warning: com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerOptions$Builder: can't find referenced class com.google.auto.value.AutoValue$Builder
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.OpenSSLProvider
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.OpenSSLProvider
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
Warning: okhttp3.internal.platform.ConscryptPlatform: can't find referenced class org.conscrypt.Conscrypt
```

The plugin doesn't look like it's including any Proguard rules, I'm guessing those would be helpful.

I'm using version 0.4.0.

SamiPatanen:
I added this in my proguard rules:
```
[...]

PR #373, opened 2018-03-19
Add consumer proguard rules to all plugins

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/369

Adds proguard rules for plugins, as well as, adds a step to the CI build to attempt to check if proguard rules work.


langsmith:
Will fix https://github.com/mapbox/mapbox-android-demo/issues/650

PR #386, opened 2018-03-28
Gestures thresholds adjustments

LukasPaczos:
Fixes issues with gestures thresholds and layer drawable refresh rates.


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -291,13 +290,6 @@ private void styleForegroundGPS(boolean isStale) {
       isStale);
   }
 
-  void updateForegroundOffset(double tilt) {
```
tobrun: was this unused? or made obsolete?

LukasPaczos: Obsolete, probably fixed in the Maps SDK core with `6.0.0`.

Issue #362, opened 2018-03-14
Attempt to remove non-JNI local reference, dumping thread logs


LukasPaczos:
We hear complaints from users about `Attempt to remove non-JNI local reference, dumping thread` logs littering the `logcat`. Referencing https://github.com/mapbox/mapbox-gl-native/issues/9688#issuecomment-373004990, we can work around those issues by using FeatureCollections and Functions/Expressions to update `LocationLayer` in various places like https://github.com/mapbox/mapbox-plugins-android/blob/1ccb76cc4b84d9e40ea1720eabf6c6a49e9e98a7/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java#L195 and https://github.com/mapbox/mapbox-plugins-android/blob/1ccb76cc4b84d9e40ea1720eabf6c6a49e9e98a7/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java#L201, instead of using layer properties setters.

cammace:
This has been quite annoying for me also, will pick this one up :)

aquabay:
I am getting this error whenever I add a ShapeSource to the map.

PR #388, opened 2018-03-28
Fix order of interpolator expression

tobrun:
This PR fixes the issue of:

>  E/mbgl: {plugins.testapp}[JNI]: Error setting property: icon-size [5]: Input/output pairs for “interpolate” expressions must be arranged with input values in strictly ascending order.


cammace:
Is this related to https://github.com/mapbox/mapbox-plugins-android/issues/362 or another issue?

LukasPaczos:
No, @cammace, only a type-error in the Expression after upgrading to the new Maps SDK.

Issue #368, opened 2018-03-15
Make all location layer attributes public


cammace:
Currently lint is throwing warnings for this

PR #392, opened 2018-03-29
Adds missing xml attributes

cammace:
closes https://github.com/mapbox/mapbox-plugins-android/issues/368

Adds new attributes and fixes private style.


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
```diff
@@ -159,7 +168,7 @@ public static LocationLayerOptions createFromAttributes(@NonNull Context context
    * automatically to their defined defaults in this library. This allows you to adjust a few
    * attributes while leaving the rest alone and maintaining their default behavior.
    *
-   * @param context your activites context used to acquire the style resource
+   * @param context your activates context used to acquire the style resource
```
danesfeder: I think you meant to fix to `activities` here?

Issue #215, opened 2017-12-11
[LocationLayerPlugin] Navigation puck chop / disappears while navigating 


danesfeder:
**Original ticket:** https://github.com/mapbox/mapbox-navigation-android/issues/581

Greetings, I have noticed a bug related to the location layer plugin while in navigation. For now at least we have determined that this bug is on all Vodafone Smart Prime 6 devices (over one thousand devices). 

At all times: the location layer does not animate fluidly on the map.

While in navigation: the location layer sometimes dissapears (it's still there if you zoom in or out) or gets left behind going off-screen even if the navigation camera animates and moves accordingly to the device's location.

I am posting a video recording of the bug, maybe it will be of more help visually in determining the problem. 

Video found [here](https://youtu.be/KHL7kfOwNac)

![whatsapp image 2017-12-10 at 22 52 15](https://user-images.githubusercontent.com/13194727/33809775-c07aec8e-de04-11e7-8c8c-292cc73afa78.jpeg)

cc @tobrun @cammace

kingfisherphuoc:
I got the same issue in `locationLayerPlugin = '0.2.0'` (snapshot version as well) and the issue only occurs when the camera `tilt > 0`. Do you have any suggestion?

lucianboboc:
+1 we're getting quite a few reports signaling this issue as well, any updates on this?

ibBogdan:
+1 i`m also having issues with this. any updates?

cammace:
Thanks for reporting everyone. @tobrun is exploring new ways to handle the navigation puck, especially when tilted, in https://github.com/mapbox/mapbox-plugins-android/pull/220. I'm moving this to the 0.4.0 milestone since the PR hasn't been completed yet.

vladgheorghe:
Hi guys, I'm sorry to insist but for my client's business this bug is currently critical due to the majority of the userbase having devices on which it reproduces.

Is there any possible workaround or quick fix to prevent this from happening until 0.4.0 is released ?

vicolmoldoveanu:
Interested about a quickfix also, we're having this issue on around ~4000 devices with a major impact on user experience.

vladgheorghe:
We noticed that the bugs reproduce at the same time on different devices. Recently, with the dynamic camera update, whenever a zoom transition occurs they snap back into position immediately. 

I uploaded 2 videos where you can see how the bug happens always in sync across devices.

Any updates on this issue ?

[...]

PR #393, opened 2018-03-29
Filter location updates and remove unused animator code

danesfeder:
Animation jank was occurring when the `LocationEngine` would push 2 or 3 updates a second (because we cancel the animation when we receive and update for a new one).

This PR filters updates, limiting to only 1 per second. 

Also removed animator code that is unused.


vicolmoldoveanu:
Hi,
Is this fix able to resolve #215  ?

LukasPaczos:
@vicolmoldoveanu this should enhance user experience even more, but please try `0.5.0-beta.1` out first, as it contains tons of fixes and adjustments one of which is a built-in camera tracking logic.

This PR will ship with the upcoming `0.5.0` release.

Issue #391, opened 2018-03-29
When camera is animating, onMoveCamera isn't called causing icon to not readjust position


cammace:
When tracking or camera is animating, the update symbol layer methods aren't being called. This is a bug downstream in Map SDK

cc: @tobrun

tobrun:
This is not a bug in gl-native as we use the `CameraUpdateFactory#moveCamera` which isn't really animating. Instead we should fix this in the plugin itself by having a callback from the animator code to:

```java
private OnCameraMoveListener onCameraMoveListener = new OnCameraMoveListener() {
 	 
    private CameraPosition lastCameraPosition;
 	 
     @Override	     @Override
     public void onCameraMove() {	     
        locationLayer.updateAccuracyRadius(getLastKnownLocation());	         
        locationLayer.updateForegroundOffset(position.tilt);
        locationLayer.updateForegroundBearing((float) position.bearing);
       }	       
     }	     
   };	   

```

PR #395, opened 2018-03-30
Invalidate onCameraMove as part of tracking animation

tobrun:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/391

cc @danesfeder @LukasPaczos

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -52,7 +52,7 @@ import static android.Manifest.permission.ACCESS_FINE_LOCATION;
  *
  * @since 0.1.0
  */
-public final class LocationLayerPlugin implements LifecycleObserver, OnCameraMoveInvalidateListener {
+public final class LocationLayerPlugin implements LifecycleObserver {
 
   private final MapboxMap mapboxMap;
   private final MapView mapView;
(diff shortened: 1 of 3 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -52,7 +52,7 @@
  *
  * @since 0.1.0
  */
-public final class LocationLayerPlugin implements LifecycleObserver {
+public final class LocationLayerPlugin implements LifecycleObserver, OnCameraMoveInvalidateListener {
```
tobrun: would be nicer to have an anonymous inner class so we don't add unnecessary public methods to the LocationLayerPlugin api

cammace: `OnCameraMoveInvalidateListener` is package private, I don't see the issue here?

LukasPaczos: @cammace When implementing an interface we are unnecessary exposing `LocationLayerPlugin#onInvalidateCameraMove`. Making it an anonymous inner class hides that method.

cammace:
I've moved the implementation into a private variable and fixed it not getting invoked inside `setBearing`. For the optimizations with onCameraMoved, I have ticketed it out and will work on it in a separate PR https://github.com/mapbox/mapbox-plugins-android/issues/413

PR #398, opened 2018-04-02
unify the latlng and bearing values

### LINKED ISSUE #394: Unify moveCamera invocations
Unify the moveCamera invocations instead of calling bearing and latlng seperatly. 

```
  private void setBearing(float bearing) {
    mapboxMap.moveCamera(CameraUpdateFactory.bearingTo(bearing));
  }

  private void setLatLng(LatLng latLng) {
    mapboxMap.moveCamera(CameraUpdateFactory.newLatLng(latLng));
  }
```

cammace:
Closes #394 

@tobrun, not sure if this is what you mean by unifying the latlng and bearing move?

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerCamera.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerCamera.java
@@ -7,6 +7,7 @@ import android.view.MotionEvent;
 import com.mapbox.android.gestures.AndroidGesturesManager;
 import com.mapbox.android.gestures.MoveGestureDetector;
 import com.mapbox.android.gestures.RotateGestureDetector;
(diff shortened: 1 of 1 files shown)


tobrun:
Yes this is what I mean with unifying though we are still updating the map twice in each animation callback while we should be able to get away with only once.

cammace:
So moving forward, I should check if the LL is in compass mode and if so use the individual update methods that are currently setup, otherwise, use a unified method which I introduced in this PR?

LukasPaczos:
Yes and besides that, we can also include some kind of a buffer for those updates, what @tobrun mentioned in https://github.com/mapbox/mapbox-plugins-android/pull/398#issuecomment-378897321. This might as well be a running runnable that updates both `LatLng` and `bearing` values from memory every couple of ms (if an update is available). The same logic can be applied to the `LocationLayer`.

LukasPaczos:
Closing in favor of https://github.com/mapbox/mapbox-plugins-android/pull/573.

PR #399, opened 2018-04-02
changed integer to dimension

cammace:
https://github.com/mapbox/mapbox-plugins-android/pull/392 accidentally was trying to get a dimension value as an integer, this PR fixes that.


LouisCAD:
Is this in a new snapshot yet?

PR #401, opened 2018-04-02
Adds text under location icon

cammace:
This allows you to place a string text under the location icon. Right now I have the demo querying the map and filtering layer features to only those with names, thus ideally, this will show you the current POI you are at. Still, need to implement navigation version of this that shows current street.

Users can use the `setPlaceText` API to set the string which is displayed.

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
@@ -2,6 +2,8 @@ package com.mapbox.mapboxsdk.plugins.testapp.activity.location;
 
 import android.annotation.SuppressLint;
 import android.content.res.Configuration;
+import android.graphics.PointF;
+import android.graphics.RectF;
 import android.location.Location;
 import android.os.Bundle;
 import android.support.annotation.VisibleForTesting;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/location/LocationLayerModesActivity.java
```diff
@@ -231,9 +239,22 @@ public void onConnected() {
...
+    PointF pointF = mapboxMap.getProjection().toScreenLocation(
+      new LatLng(location.getLatitude(), location.getLongitude()));
+
+    // Add padding to the point
+    RectF rectF = new RectF(pointF.x + 50, pointF.y + 50, pointF.x + 50, pointF.y + 50);
+
+    List<Feature> features = mapboxMap.queryRenderedFeatures(rectF, has("name"));
+    if (!features.isEmpty()) {
+      locationLayerPlugin.setPlaceText(features.get(0).getStringProperty("name"));
+    }
```
tobrun: While I like the idea shown above, I wonder if we need it as default behaviour?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -239,6 +239,10 @@ public void applyStyle(LocationLayerOptions options) {
     updateMapWithOptions(options);
   }
 
+  public void setPlaceText(String placeText) {
```
tobrun: missing javadoc

PR #402, opened 2018-04-02
removes duplicate APIs

### LINKED ISSUE #400: Remove LocationLayerOption APIs from LocationLayerPlugin
We currently expose setters for 3 properties which can also be set through the LocationLayerOptions class. These are the APIs: https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java#L260-L283

cc: @tobrun @LukasPaczos @danesfeder

cammace:
Closes #400

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -236,6 +236,7 @@ public final class LocationLayerPlugin implements LifecycleObserver {
       staleStateManager.onStop();
     }
     staleStateManager.setDelayTime(options.staleStateTimeout());
+    updateMapWithOptions(options);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -526,19 +481,6 @@ private void disableLocationLayerPlugin() {
     locationLayer.hide();
   }
 
-  private void updateMapWithOptions(LocationLayerOptions options) {
```
LukasPaczos: We cannot simply remove this method. We need to call it in `LocationLayerPlugin#applyStyle` or move its logic there. Otherwise, those values won't be set on the map.

cammace: Good catch, I readded the `updateMapWithOptions` method but without the checks since inside LocationLayerOptions we check the values.

PR #403, opened 2018-04-02
Remove and readd onMapChange in onStart/onStop

### LINKED ISSUE #363: Remove map change listener when locationlayer isn't being used.

cammace:
Closes #363 

When the location layer plugin's disabled, we remove the onMapChange listener.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -416,13 +416,30 @@ public final class LocationLayerPlugin implements LifecycleObserver {
   @RequiresPermission(anyOf = {ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION})
   @OnLifecycleEvent(Lifecycle.Event.ON_START)
   public void onStart() {
+    onLocationLayerStart();
+    if (isEnabled) {
+      mapView.addOnMapChangedListener(onMapChangedListener);
+    }
+  }
+
+  /**
+   * Required to place inside your activities {@code onStop} method.
+   *
+   * @since 0.1.0
+   */
+  @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+  public void onStop() {
+    onLocationLayerStop();
+    mapView.removeOnMapChangedListener(onMapChangedListener);
+  }
+
+  void onLocationLayerStart() {
     if (isEnabled) {
       if (locationEngine != null) {
         locationEngine.addLocationEngineListener(locationEngineListener);
       }
       setLastLocation();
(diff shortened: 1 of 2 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -489,6 +490,7 @@ public void onStop() {
     if (mapboxMap != null) {
       mapboxMap.removeOnCameraMoveListener(onCameraMoveListener);
     }
+    mapView.removeOnMapChangedListener(onMapChangedListener);
```
LukasPaczos: Won't this one unregister the listener during style switch resulting in disabling the location layer? See https://github.com/mapbox/mapbox-plugins-android/blob/86a67081e4a188433cfd7f51740c9dfa2fafc43e/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java#L632-L641.

tobrun: that only happens with onMapReady callbacks not with end user provided map change listeners

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -601,13 +612,13 @@ public void onStaleStateChange(boolean isStale) {
     @Override
     public void onMapChanged(int change) {
       if (change == MapView.WILL_START_LOADING_MAP) {
-        onStop();
+        onLocationLayerStop();
       } else if (change == MapView.DID_FINISH_LOADING_STYLE) {
         locationLayer.initializeComponents(options);
         locationLayerCamera.initializeOptions(options);
         setRenderMode(locationLayer.getRenderMode());
```
LukasPaczos: If the plugin is disabled and we change the style this line will draw the puck in the last location even though the plugin is disabled.

cammace: Yes, want to make this clear that this is a bug that is in master though, not something I introduced in this PR...

cammace: After looking into this, I'm not finding a solution to this fix. Will open a ticket and revisit in a separate PR.

cammace: https://github.com/mapbox/mapbox-plugins-android/issues/515

PR #411, opened 2018-04-04
Fix wrong resource type expectation

LouisCAD:
Was trying to get a dimension resource as an integer resource.
This fixes these exceptions that occur at runtime:
java.lang.UnsupportedOperationException: Can't convert value at index 22 to integer: type=0x5


LouisCAD:
Oops. Duplicate of #399

cammace:
Nevertheless, thanks for contributing! Silly mistake on my part. Going ahead and closing this PR in favor of #399.

Issue #413, opened 2018-04-04
Stop making layer unnecissary updates when onCameraMove is invoked.


cammace:
Capturing from https://github.com/mapbox/mapbox-plugins-android/pull/395:

> Additionally, it'd be great to optimize LocationLayerPlugin#onCameraMove not to update when unnecessary, for ex. we don't need to update accuracy radius when zoom has not changed etc. This might help with table overflow issues we've seen recently.

cc: @LukasPaczos @tobrun

PR #426, opened 2018-04-16
Use json features instead of layer properties setters

LukasPaczos:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/362 and closes https://github.com/mapbox/mapbox-plugins-android/issues/413.


LukasPaczos:
@tobrun Performance feels alright, didn't compare FPS against old implementation but that's a good point, I'll do that.
`Attempt to remove non-JNI local reference` are gone unless you change the render mode because we need to adjust layer visibility then, and that needs to be set, we can't use `Expression`s.

PR #427, opened 2018-04-16
Initialize in a stale state

LukasPaczos:
I think it improves UX if we initialize in a stale state and dismiss that with the first location update.


cammace:
Agreed, we also need to make sure we keep the stale state consistent when location layer modes are changed (from tracking to compass for example). Will ticket this out.

PR #430, opened 2018-04-17
Max animation duration when tracking

LukasPaczos:
Added checks for maximum animation duration time between location samples (2 seconds) as well as fixed animation duration when switching between camera modes (750 ms).

/cc @tobrun @cammace


cammace:
Going ahead and merging.

PR #431, opened 2018-04-17
Remove invalid Location update check

danesfeder:
We found that the invalid update check was causing the initial delay when showing the user's location with the `LocationLayerPlugin`.  

After further testing, it doesn't seem the check is necessary as `Location` APIs should not be sending the same update in less that a seconds time.


danesfeder:
@LukasPaczos yep, good catch - updated

LukasPaczos:
Also noting, that users should default to `LocationEngine#setFastestInterval` and `LocationEngine#setInterval` to manage location updates frequency.

PR #455, opened 2018-04-22
Fix expressions not localizing

### LINKED ISSUE #374: Localization plugin should transform expressions
The map localization plugin added in mapbox/mapbox-plugins-android#74 only works if a `text-field` property is set to a constant value that uses the legacy token syntax, e.g. `{name_en}`. In a style such as Mapbox Streets, some layers’ `text-field`s are set to camera functions, so they remain unlocalized. To make matters worse, future versions of these and other styles will be written for versions of the SDK that support expressions. `{name_en}` will become `["get", "name_en"]`, so this plugin will end up having no effect whatsoever.

/ref mapbox/mapbox-gl-native#10713 mapbox/mapbox-gl-js#6197
/cc @langsmith @tobrun

cammace:
Closes #374 

Currently blocked by the issues mentioned in the issue.


tobrun:
done in https://github.com/mapbox/mapbox-plugins-android/pull/654

PR #463, opened 2018-04-24
adds transifex

cammace:


1ec5:
Also, consider adding something to the contributing guide about adding translations, so that we don’t end up getting PRs containing translations that aren’t in Transifex.

Issue #470, opened 2018-04-30
LocationLayerPlugin not visible after setting mapStyleUrl or random crashes


vladgheorghe:
Hello, in onMapReady I am initializing the locationlayerplugin. Immediately after I am setting the map style through the method map.setMapStyleUrl. The plugin is not visible and sometimes we are experiencing crashes immediately after the callback onStyleLoaded gets called. Reproduces always on 0.5.0 with maps sdk 6.0.1

cammace:
I'll add a test case for this. In https://github.com/mapbox/mapbox-plugins-android/blob/master/app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerTest.kt#L166, we are setting the location layer plugin up first and then changing the map style. Opposite of what you're doing.

cammace:
I wrote a new Isolated test class for the Location Layer Plugin which just uses a default `SupportMapFragment` as the view. then, I went through the steps to try reproducing your issue but it seems no matter where I change the map style, the Location Layer seems to handle the change fairly well. You can see the test in https://github.com/mapbox/mapbox-plugins-android/pull/471. Feel free to comment on this issue or the PR if I got something wrong. 

Is there a reason for changing the map style immediately when the mapReady callback is invoked? It's better to initialize your map style through XML or MapboxMapOptions if you can that way when the maps initially rendered it uses your style rather than downloading the default Street style first and then switching to your desired style.

cammace:
Since I wasn't able to reproduce this issue in #470, I am going ahead and closing this ticket. Will reopen if you are still running into issues and my test wasn't testing your use case.

vladgheorghe:
Unfortunately, it seems that I am experiencing a different behaviour. I'm posting the code I am using and an attachment of the bug reproducing.

```
public void onMapReady(MapboxMap mapboxMap) {
        map = mapboxMap;
        map.getUiSettings().setTiltGesturesEnabled(false);
        initCamera();
        initLocationLayer();
        //this gets based on user prefences
       map.setStyleUrl(MAPBOX_DAY_STYLE_URL,  MainActivity.this);

    @SuppressWarnings({"MissingPermission"})
    private void initLocationLayer() {
        LocationLayerOptions options = LocationLayerOptions.builder(this)
                .padding(mapViewPadding)
                .bearingDrawable(R.drawable.icon_bearing)
                .gpsDrawable(R.drawable.icon_user_puck)
                .build();
[...]

PR #483, opened 2018-05-04
fix source missing when styles reloaded

cammace:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/470

Still results in an issue when the user locations null still -> map style changes -> source gets added and location icons shown at "null island" (coordinates 0,0). Will look into fixing this before merging.


cammace:
I've reopened the original ticket which brings back the null island issue but IMO, it's a much better issue then the location source not getting added correctly. Therefore, this PR has reverted back to the original logic for the 0.5.1 release and I'll look more into it after the release in a separate PR.

PR #520, opened 2018-05-18
Update LocationLayerPlugin permission annotations

danesfeder:
Underlying code working with `LocationEngine` in the `LocationLayerPlugin` was updated, but the methods annotating the required permissions for certain methods were not.  These changes rearrange the annotations to reflect the new code changes.


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -433,35 +430,6 @@ public void onStop() {
     mapView.removeOnMapChangedListener(onMapChangedListener);
   }
 
-  void onLocationLayerStart() {
```
cammace: Any reason for changing the access modifiers here? Changing to private causes a synthetic accessor method to be created. In my opinion, it should be the opposite, we should prevent synthetic accessor methods from being created everywhere possible as this adds to the method count.

danesfeder:
@cammace will the permission annotations in this PR conflict with the engine work being re-added in #527?

PR #526, opened 2018-05-21
Fix accuracy circle color not being set correctly

### LINKED ISSUE #482: Location Layer accuracy color not being set correctly

cammace:
Closes #482


LukasPaczos:
Android color ints are not supported in Expressions so we need to convert them to RGBA strings. Try
```
locationFeature.addStringProperty(PROPERTY_ACCURACY_COLOR, PropertyFactory.colorToRgbaString(accuracyColor));
``` 
instead of [pushing the Android `int`](https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java#L360) @cammace.

Issue #510, opened 2018-05-14
Not seeing location layer marker after upgrading mapbox


stoefln:
After upgrading to MapBox 6.01 and Location Layer Plugin 0.5.2 I don't see a marker anymore.
I figured out that adding a LocationEngine solves the issue, however we have a complicated architecture which includes a LocationService, which makes using the LocationEngine more complicated. Do I really need a LocationEngine and when did that requirement change?

cammace:
Hey there @stoefln, first, to address your versioning. If you are including the Location Layer Plugin inside your app as a dependency, you no longer need to also include the Mapbox Map SDK dependency since the Location Layer offers the SDK and correct version of it transitively. This might be the issue with not seeing the Location Layer.

> Do I really need a LocationEngine and when did that requirement change?

No, you shouldn't need to provide a LocationEngine when using the Location Layer Plugin. Being able to manually update the location layer is a designed feature specifically for use cases like yours, this is why the param is annotated `nullable`. Could you provide snippets of code that I can use to reproduce the issue, specifically around initializing the Location Layer and manually providing updates.

cammace:
Closing for now but can revisit once more information is provided.

cc: @zugaldia

stoefln:
@cammace I cleaned my code so I can figure out why it's not working without locationEngine. Now its really simple and it's basically just:
```
          locationEngine = new 
 LocationEngineProvider(getContext()).obtainBestLocationEngineAvailable();
            locationEngine.setPriority(LocationEnginePriority.HIGH_ACCURACY);
            locationEngine.activate();
            locationEngine.addLocationEngineListener(new LocationEngineListener() {
                @Override
                public void onConnected() {
                    Timber.d("xxx onConnected");
                    locationEngine.requestLocationUpdates();
                }

                @Override
                public void onLocationChanged(Location location) {
                    Timber.d("xxx onLocationChanged");
                    originLocation = location;
                }
            });
// with locationEngine, everything works. not providing does not work
            locationPlugin = new LocationLayerPlugin(mapView, getMapboxMap(), null);
            locationPlugin.setLocationLayerEnabled(true);
            locationPlugin.setCameraMode(CameraMode.TRACKING_GPS);
            locationPlugin.setRenderMode(RenderMode.GPS);

            locationPlugin.onStart();
```

danesfeder:
[...]

PR #539, opened 2018-05-24
Initialize LocationEngine for all constructors if null

danesfeder:
Re: https://github.com/mapbox/mapbox-plugins-android/issues/510#issuecomment-391717083

If we only want the constructor added in #527 to initialize a default `LocationEngine`, then we should mark the other constructor `LocationEngine` params as `@NonNull` currently marked as nullable. 

Happy to change the annotations or use the code I submitted originally


LukasPaczos:
The nullable `LocationEngine` option is there for users how would like to rely on pushing location updates via the force option, without any kind of `LocationEngine` at all. Personally I feel like it needs to stay, but maybe we could adjust the docs to highlight possible scenarios?

danesfeder:
Ah yeah @LukasPaczos, thanks for the clarification there.  Yeah an update to the docs specifying this should help

Issue #538, opened 2018-05-24
Suggestion: make CameraMode constants easier to work with


kjkrum:
As the location layer plugin is a 0.x version, are the maintainers open to breaking changes?

I find the `CameraMode` constants a little messy to work with. I wrote a bunch of ugly code tonight that looks like this:

```java
mFollowToggle.setOnClickListener(view -> {
	switch(mLocationLayer.getCameraMode()) {
		case NONE:
			mLocationLayer.setCameraMode(TRACKING);
			break;
		case TRACKING:
			mLocationLayer.setCameraMode(NONE);
			break;
		case NONE_COMPASS:
			mLocationLayer.setCameraMode(TRACKING_COMPASS);
			break;
		case TRACKING_COMPASS:
			mLocationLayer.setCameraMode(NONE_COMPASS);
			break;
	}
});
```
My first thought was that it might be nicer if the constants were xor'able bit flags, but I'm not sure that's feasible since, for example, there's more than one mode for bearing tracking. So maybe `CameraMode` should be an enum implementing an interface with various set and toggle methods.

I'll put more thought into it and maybe send a PRQ if this sounds like something the team would consider.

cammace:
Agree that the constants should be bit flags, I've dropped some comments on your PR which made the suggested changes. The nice thing is that from the PR, it looks like the changes you made only affect the internal code and doesn't change any public APIs, thus, no breaking changes 🎉

kjkrum:
@cammace I haven't forgotten about this. I've just been sick.

kjkrum:
@cammace I'm not sure what you mean about this only affecting internal code. Is `LocationLayerPlugin.setCameraMode(int)` not considered part of the public API?

zugaldia:
I'm gonna close this ticket as per https://github.com/mapbox/mapbox-plugins-android/issues/684#issuecomment-433947880 the functionality provided by the Location Layer Plugin is now part of the core Maps SDK (there's no need to add additional plugins) and the plugin is now deprecated. If this ticket is still relevant for the new `LocationComponent`, please cut a new ticket on https://github.com/mapbox/mapbox-gl-native/issues/new?template=android.md. Thanks!

PR #542, opened 2018-05-26
Changed camera mode constants to bit flags. #538

kjkrum:
Added `FOLLOW_xxx` and `ORIENT_xxx` constants that are bit flags, and redefined the values of existing constants in terms of these. While not bytecode compatible, this is source code compatible with normal usage of the existing constants.


zugaldia:
I'm gonna close this PR as per https://github.com/mapbox/mapbox-plugins-android/issues/684#issuecomment-433947880 the functionality provided by the Location Layer Plugin is now part of the core Maps SDK (there's no need to add additional plugins) and the plugin is now deprecated. If this ticket is still relevant for the new `LocationComponent`, please cut a new PR on https://github.com/mapbox/mapbox-gl-native. Thanks!

PR #554, opened 2018-06-13
Problem with option 'country'

gustron12:
When i try to search with country option, give me the Offline error, I see that this line isn't correct


LukasPaczos:
Great catch @gustron12!
Closing in favor of https://github.com/mapbox/mapbox-plugins-android/pull/555 as CI won't run for external PRs.

PR #572, opened 2018-07-04
Showing accuracy for forced location

### LINKED ISSUE #469: LocationLayerPlugin 0.5 not having anymore accuracy animation
In new 0.5 location plugin there is accuracy circle but no animation it looks bad in my opinion. Can you add animation or add api to disable accuracy circle at all.  Also, how to change gps marker for location layer, with png no vector?

LukasPaczos:
Refs #469.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -20,6 +20,7 @@ import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.maps.MapView;
 import com.mapbox.mapboxsdk.maps.MapView.OnMapChangedListener;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
+import com.mapbox.mapboxsdk.maps.MapboxMap.OnCameraIdleListener;
 import com.mapbox.mapboxsdk.maps.MapboxMap.OnCameraMoveListener;
 import com.mapbox.mapboxsdk.maps.MapboxMap.OnMapClickListener;
 import com.mapbox.mapboxsdk.plugins.locationlayer.modes.CameraMode;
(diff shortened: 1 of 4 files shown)


LukasPaczos:
I initially misinterpreted #469 - it points out that accuracy circle doesn't animate when the new location (with different accuracy than previous) is provided, it just jumps to the required value.

https://github.com/mapbox/mapbox-plugins-android/issues/469#issuecomment-399092359 still stands and is fixed by this PR.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -358,7 +364,11 @@ public LocationEngine getLocationEngine() {
   @Nullable
   @RequiresPermission(anyOf = {ACCESS_FINE_LOCATION, ACCESS_COARSE_LOCATION})
   public Location getLastKnownLocation() {
-    return locationEngine != null ? locationEngine.getLastLocation() : null;
+    Location location = locationEngine != null ? locationEngine.getLastLocation() : null;
+    if (location == null) {
+      location = lastLocation;
```
danesfeder: @LukasPaczos Per javadoc above - here we are trying to get the last known location from the location engine and if nothing is found, using `lastLocation` which could also possibly be null?  Just clarifying here, I think it makes sense to add the extra check in case a developer was forcing `Location` updates.

LukasPaczos: Yes, `lastLocation` can also be null if no updates have been pushed to the plugin yet. This is our last resort if `locationEngine` is `null` or fails to provide `non-null` location, like a scenario when we are only forcing location updates.

The idea here is, that `LocationEngine` can be in a position to provide last location (or just more recent location) even if no location updates have been pushed to the plugin itself.

LukasPaczos:
> you're saying the animation from different accuracies is fixed

Unfortunately, no. This is something I'm working on right now. This PR fixes an issue which occurred when you didn't use `locationEngine` (only forcing location updates) and zoomed in/out - then accuracy would be gone. We are using `getLastKnownLocation` to recalculate the radius for a new zoom level, and it would always return `null`, because there was no engine. Now we have `lastLocation` variable to fall back to.

Issue #567, opened 2018-07-03
Option to disable animation


LukasPaczos:
We are hearing voices that smooth tracking animation puts a real stress on the battery consumption, CPU usage and introduces a delay. In order to improve the experience on lower-end devices, we could expose an option to disable the smooth animation.

/cc @danesfeder

zugaldia:
Issue moved to [mapbox/mapbox-gl-native #13225](https://github.com/mapbox/mapbox-gl-native/issues/13225) via [**ZenHub**](https://www.zenhub.com/)

PR #573, opened 2018-07-05
Unify bearing and target CameraUpdate

### LINKED ISSUE #394: Unify moveCamera invocations

danesfeder:
Closes #394 

This PR seeks to update the `MapboxMap` camera `float` bearing and `LatLng` target at the same time.  It also introduces a `LocationLayerOption#refreshIntervalInMillis()` to allow customization of how often you'd like the camera to update.  

@LukasPaczos @tobrun if this approach seems okay, I can use the same runnable with the layer, so both the camera and layer updates will reflect the interval provided by the options. 

TODO:
- [x] Updated docs

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
@@ -75,6 +75,11 @@ public abstract class LocationLayerOptions implements Parcelable {
   private static final long STALE_STATE_DELAY_MS = 30000;
 
   /**
+   * By default, the location layer and camera will update at 60/fps
+   */
+  private static final double REFRESH_INTERVAL_IN_MILLIS_DEFAULT = 16.67;
+
+  /**
    * Construct a new Location Layer Options class using the attributes found within a style
    * resource. It's important to note that you only need to define the attributes you plan to
    * change and can safely ignore the other attributes which will be set to their default value.
@@ -220,7 +225,8 @@ public abstract class LocationLayerOptions implements Parcelable {
       .minZoom(MIN_ZOOM_DEFAULT)
       .maxZoomIconScale(MAX_ZOOM_ICON_SCALE_DEFAULT)
       .minZoomIconScale(MIN_ZOOM_ICON_SCALE_DEFAULT)
-      .padding(PADDING_DEFAULT);
+      .padding(PADDING_DEFAULT)
+      .refreshIntervalInMillis(REFRESH_INTERVAL_IN_MILLIS_DEFAULT);
   }
 
   /**
@@ -548,6 +554,15 @@ public abstract class LocationLayerOptions implements Parcelable {
    * @since 0.5.0
    */
   public abstract float trackingMultiFingerMoveThreshold();
+
+  /**
+   * Determines how often the {@link LocationLayer} and {@link LocationLayerCamera}
+   * will update.  By default this is set to 60 frames per second.
+   *
+   * @return refresh interval in milliseconds
+   * @since 0.6.0
+   */
+  public abstract double refreshIntervalInMillis();
 
   /**
    * Builder class for constructing a new instance of {@link LocationLayerOptions}.
@@ -890,6 +905,15 @@ public abstract class LocationLayerOptions implements Parcelable {
      * @since 0.5.0
      */
     public abstract Builder trackingMultiFingerMoveThreshold(float moveThreshold);
+
+    /**
+     * Determines how often the {@link LocationLayer} and {@link LocationLayerCamera}
+     * will update.  By default this is set to 60 frames per second.
+     *
+     * @param refreshIntervalInMillis interval in milliseconds
+     * @since 0.6.0
+     */
+    public abstract Builder refreshIntervalInMillis(double refreshIntervalInMillis);
 
     abstract LocationLayerOptions autoBuild();
 
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -47,7 +47,7 @@ import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.
  * display a larger icon (customized with {@link LocationLayerOptions#gpsDrawable()}) we call the user puck.
  * <p>
  * This plugin also offers the ability to set a map camera behavior for tracking the user
- * location. These different {@link CameraMode}s will track, stop tracking the location based on the
+ * location. These different {@link CameraMode}s will track or cancel tracking the location based on the
  * mode set with {@link LocationLayerPlugin#setCameraMode(int)}.
  * <p>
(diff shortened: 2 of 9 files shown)


danesfeder:
This may also address https://github.com/mapbox/mapbox-plugins-android/issues/567, as you'll be able to essentially disable interpolation by increasing the refresh interval.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -513,8 +513,12 @@ private void initialize() {
     mapboxMap.addOnMapLongClickListener(onMapLongClickListener);
 
     locationLayer = new LocationLayer(mapView, mapboxMap, options);
-    locationLayerCamera = new LocationLayerCamera(
-      mapView.getContext(), mapboxMap, cameraTrackingChangedListener, options, onCameraMoveInvalidateListener);
+
+    CameraUpdateProvider cameraUpdateProvider = new CameraUpdateProvider(mapboxMap,
```
LukasPaczos: This provider's queue needs to be cleared similarly to `StaleStateManager#onStop`.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CameraUpdateProvider.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  void pushLatLng(LatLng target) {
+    this.target = target;
+  }
+
+  private CameraUpdate buildCameraUpdate() {
+    CameraPosition cameraPosition = mapboxMap.getCameraPosition();
+    CameraPosition.Builder cameraPositionBuilder = new CameraPosition.Builder();
+    if (bearing > 0 && bearing != cameraPosition.bearing) {
```
LukasPaczos: I think our current logic doesn't ensure that the target bearing is always 0-360.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
```diff
@@ -73,6 +73,9 @@
    */
   private static final long STALE_STATE_DELAY_MS = 30000;
 
+  // TODO docs
+  private static final int REFRESH_INTERVAL_IN_MILLIS_DEFAULT = 1;
```
LukasPaczos: I think we can make the default refresh rate of 60FPS? That'd require a delay of 16.67 ms. Let's test this out though and decide more experimentally.

danesfeder: Sounds good, I'll update the variable and we can test it out

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CameraUpdateProvider.java
danesfeder: Does the shortest rotation util address this? https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/Utils.java#L32-L40

danesfeder:
@LukasPaczos addressed your comments and asked some follow up questions.  Test for the camera provider / docs for the options were also added.  

Now working with both the camera and layer (here at 1/fps):
ezgif com-video-to-gif

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CameraUpdateProvider.java
LukasPaczos: I don't think that's the case if the previous and current bearings are >360 or <-360, and when both are < 0. We probably should update the utils method to better normalize the bearing. Also, the bearing can be equal to 0.

LukasPaczos: Additionally, we need to check here if we actually want to modify camera's bearing, otherwise in a mode like `CameraModer.TRACKING`, where we are tracking only target, we are unable to rotate the camera.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CameraUpdateProvider.java
```diff
@@ -0,0 +1,64 @@
...
+      cameraPositionBuilder.bearing(bearing);
+    }
+    if (target != null && !target.equals(cameraPosition.target)) {
```
LukasPaczos: Similarly, as in https://github.com/mapbox/mapbox-plugins-android/pull/573/files#r200948002, we need to check if we are actually tracking the position and not only bearing, otherwise, we won't be able to move the camera.

LukasPaczos: Maybe we can move those checks out of [LocationLayerCamera](https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerCamera.java#L120-L133) and in here?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CameraUpdateProvider.java
```diff
@@ -0,0 +1,64 @@
+package com.mapbox.mapboxsdk.plugins.locationlayer;
+
+import com.mapbox.mapboxsdk.camera.CameraPosition;
+import com.mapbox.mapboxsdk.camera.CameraUpdate;
+import com.mapbox.mapboxsdk.camera.CameraUpdateFactory;
+import com.mapbox.mapboxsdk.geometry.LatLng;
+import com.mapbox.mapboxsdk.maps.MapboxMap;
+
+class CameraUpdateProvider implements RefreshPluginRunnable.RefreshPluginListener {
```
LukasPaczos: Generally, what about implementing the `RefreshPluginRunnable.RefreshPluginListener` in the `LocationLayerCamera`? That would simplify the checks needed for the animation and reflect the setup that we have with the `LocationLayer`.

zugaldia:
I'm gonna close this PR as per https://github.com/mapbox/mapbox-plugins-android/issues/684#issuecomment-433947880 the functionality provided by the Location Layer Plugin is now part of the core Maps SDK (there's no need to add additional plugins) and the plugin is now deprecated. If this ticket is still relevant for the new `LocationComponent`, please cut a new PR on https://github.com/mapbox/mapbox-gl-native. Thanks!

PR #574, opened 2018-07-05
Allow pre-loaded MapboxMap images / maki icons in LocationLayerOptions

### LINKED ISSUE #486: Support usage of images already loaded in map style for location layer drawables
Pointed out in https://github.com/mapbox/mapbox-plugins-android/issues/183#issuecomment-356668734, it makes sense to also allow users to use the bitmap id which defines the sprite inside the map style for their location layer icons.

danesfeder:
Closes #486 

This PR introduces options to use `String` names for icons that have already been added to the `MapboxMap` with `MapboxMap#addImage`.  Developers may also provide names for [Maki icons](https://www.mapbox.com/maki-icons/) that are already loaded in their given style. 

device-2018-07-05-155217

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
@@ -41,8 +41,13 @@ import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_ACCURACY_ALPHA;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_ACCURACY_COLOR;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_ACCURACY_RADIUS;
+import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_BACKGROUND_ICON;
+import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_BACKGROUND_STALE_ICON;
+import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_BEARING_ICON;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_COMPASS_BEARING;
+import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_FOREGROUND_ICON;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_FOREGROUND_ICON_OFFSET;
+import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_FOREGROUND_STALE_ICON;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_GPS_BEARING;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_LOCATION_STALE;
 import static com.mapbox.mapboxsdk.plugins.locationlayer.LocationLayerConstants.PROPERTY_SHADOW_ICON_OFFSET;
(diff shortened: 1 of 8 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -190,15 +190,21 @@ private void setLayerVisibility(String layerId, boolean visible) {
...
     addAccuracyLayer();
   }
 
-  private void addSymbolLayer(String layerId, String beforeLayerId) {
+  private void addSymbolLayer(String layerId, String beforeLayerId, LocationLayerOptions options) {
+    String foregroundIconString = buildIconString(options.foregroundName(), FOREGROUND_ICON);
+    String foregroundStaleIconString = buildIconString(options.foregroundStaleName(), FOREGROUND_STALE_ICON);
+    String backgroundIconString = buildIconString(options.backgroundName(), BACKGROUND_ICON);
+    String backgroundStaleIconString = buildIconString(options.backgroundStaleName(), BACKGROUND_STALE_ICON);
+    String bearingIconString = buildIconString(options.bearingName(), BEARING_ICON);
```
LukasPaczos: I think this values would have to be moved to the location feature, otherwise, you'd not be able to revert this custom image names when resetting the style on an already initialized plugin.

We could also document this behavior and leave as is, thoughts?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -338,53 +351,56 @@ private void setLocationPoint(Point locationPoint) {
   // Styling
   //
 
+  private void styleForeground(@NonNull LocationLayerOptions options) {
+    boolean isGpsRenderMode = renderMode == RenderMode.GPS;
+    if (options.foregroundName() == null) {
```
LukasPaczos: Can we have separate cases for `foregroundName` and `foregroundGpsName`? This would allow for providing an id for both normal foreground icon, as well as a GPS one.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -338,53 +351,56 @@ private void setLocationPoint(Point locationPoint) {
...
+  private void styleBearing(LocationLayerOptions options) {
+    if (options.bearingName() == null) {
+      Drawable bearingDrawable = getDrawable(context, options.bearingDrawable(), options.bearingTintColor());
+      mapboxMap.addImage(BEARING_ICON, getBitmapFromDrawable(bearingDrawable));
+    }
   }
 
   private void styleShadow(LocationLayerOptions options) {
     Drawable shadowDrawable = ContextCompat.getDrawable(context, R.drawable.mapbox_user_icon_shadow);
+    if (shadowDrawable == null) {
```
LukasPaczos: This drawable is provided by the plugin, can it be `null`?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
```diff
@@ -192,15 +192,22 @@ private void setLayerVisibility(String layerId, boolean visible) {
...
+    addSymbolLayer(SHADOW_LAYER, BACKGROUND_LAYER, options);
+    addSymbolLayer(BACKGROUND_LAYER, FOREGROUND_LAYER, options);
+    addSymbolLayer(FOREGROUND_LAYER, null, options);
+    addSymbolLayer(BEARING_LAYER, null, options);
     addAccuracyLayer();
   }
 
-  private void addSymbolLayer(String layerId, String beforeLayerId) {
+  private void addSymbolLayer(String layerId, String beforeLayerId, LocationLayerOptions options) {
+    String foregroundString = options.gpsName() != null ? options.gpsName() : options.foregroundName();
```
danesfeder: @LukasPaczos ran into a possible issue here - the GPS / Foreground drawables both share the foreground icon layer.  At the time we add the symbol layer here, we don't know what render mode we are in yet.  So a decision needs to be made as far as what String is used.  Here I prioritized the GPS name.  What do you think?

danesfeder:
@LukasPaczos thanks for the feedback!  Any ideas on how to alternate between `String` icon names and `Drawable`s from the options?  Are we forced to reload the map here?

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
LukasPaczos: I think the resolution here would be keeping the string IDs of the used images in the `locationFeature`, this way we would be able to update them in the runtime, without reloading the map.

danesfeder:
@LukasPaczos the `Feature` changes / tests you added look 💯 ✅

PR #577, opened 2018-07-09
Animate accuracy ring when new value is provided

### LINKED ISSUE #469: LocationLayerPlugin 0.5 not having anymore accuracy animation

LukasPaczos:
Closes #469.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
@@ -10,12 +10,14 @@ import android.view.animation.LinearInterpolator;
 
 import com.mapbox.mapboxsdk.camera.CameraPosition;
 import com.mapbox.mapboxsdk.geometry.LatLng;
+import com.mapbox.mapboxsdk.plugins.locationlayer.camera.AccuracyAnimator;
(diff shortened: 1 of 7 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerAnimator.java
```diff
@@ -304,6 +343,21 @@ private void playCompassAnimators(long duration) {
...
+  private void updateAccuracyAnimators(float targetAccuracyRadius, float previousAccuracyRadius) {
+    cancelAccuracyRadiusAnimations();
+    accuracyRadiusAnimator = new AccuracyAnimator(previousAccuracyRadius, targetAccuracyRadius);
+    accuracyRadiusAnimator.addUpdateListener(accuracyRadiusUpdateListener);
+  }
+
+  private void playAccuracyRadiusAnimators(long duration) {
+    List<Animator> accuracyAnimators = new ArrayList<>();
+    accuracyAnimators.add(accuracyRadiusAnimator);
+    AnimatorSet accuracyAnimatorSet = new AnimatorSet();
```
danesfeder: @LukasPaczos do we need to create an `AnimatorSet` here if the `accuracyRadiusAnimator` is the only one being added?

Issue #544, opened 2018-05-30
Setting cameraMode cancels animation


stoefln:
When switching from overview to navigation mode in my app, I need to tilt the camera (3d perspective when driving). Additionally we need to set the cameraMode to a different one. The problem is that the tilt animation gets canceled when setting the cameraMode. I tried to do one after the other, using the animation callback, but its not a working solution either: when the user switches between modes quickly, the cameraMode cancels even the following animation. Blocking the user for around a second to quickly switch back to the first mode works, but it's not providing a good UX. 
Is there an easier way to do this / am I missing something?
It seems to me that setting cameraMode is an asynchronous task, which takes some time. In this case there should be a callback available IMO. 

A similar issue is that the animation gets cancelled on location update. How can this be avoided?

LukasPaczos:
In order to mix-up camera tracking with any other type of animation at the same time, you'll have to implement `AndroidAnimators` value animation and push updates through `MapboxMap#moveCamera`. We will try to make this easier by updating the camera tracking API itself or taking a different approach with the [Camera Animator Plugin](https://github.com/mapbox/mapbox-plugins-android/issues/550).

1ec5:
> updating the camera tracking API itself

mapbox/mapbox-gl-native#3625

LukasPaczos:
Clearing up the issue - currently, the developer is unable to use Mapbox's camera API to move, zoom or tilt the camera while the plugin's camera tracking is engaged because it will cancel those animations.

Changing the camera position can be easily worked around by disengaging the tracking and working with `MapboxMap#CancelableCallback` to synchronize the process.

However, there is no possibility to zoom or tilt the camera while tracking without disrupting the UX.

Beside https://github.com/mapbox/mapbox-plugins-android/issues/550, another approach to solve this issue would be introducing additional methods to the plugin that would zoom in/out and tilt the camera using `AndroidAnimators`. This won't raise concerns from https://github.com/mapbox/mapbox-plugins-android/issues/550#issuecomment-394807429 as we won't expose ways to manipulate camera's target. This would serve as a temporary workaround until there is a concrete resolution for https://github.com/mapbox/mapbox-gl-native/issues/3625.

/cc @danesfeder @kkaefer @julianrex @tobrun

stoefln:
I know this is an old ticket, but I am again facing the same issue with the newest version of mapbox and the navigation sdk.
[...]

PR #583, opened 2018-07-16
Additional camera zoom/tilt methods

LukasPaczos:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/544.

While working on https://github.com/mapbox/mapbox-plugins-android/pull/577 I noticed that it's hard to extend the `LocationLayerAnimator` with new animation types, therefore, I refactored the code and updated the naming schema. Other than that, in this PR:
- https://github.com/mapbox/mapbox-plugins-android/commit/ff8872cddb5fb537cbfe131be82ca290decff098 fixes an issue where the fixed focal point would persist if camera modes where switched in a quick succession.
- Added new methods that allow zooming and tilting the camera while any of the camera modes is engaged, besides the `CameraMode#NONE`.

ezgif com-video-to-gif 2

I'm open for discussion whether the approach taken with the methods that work only if the camera mode is engaged should be the way to go. The example of the required implementation can be found in the [test app changes](https://github.com/mapbox/mapbox-plugins-android/compare/544-additional-camera-methods?expand=1#diff-fb25a5bc49f4cfbd354697ccf7831b29R320).


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/PluginAnimatorCoordinator.java
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    float previousAccuracyRadius = getPreviousAccuracyRadius();
+    updateAccuracyAnimators(targetAccuracyRadius, previousAccuracyRadius);
+    playAccuracyAnimator(noAnimation ? 0 : ACCURACY_RADIUS_ANIMATION_DURATION);
+
+    this.previousAccuracyRadius = targetAccuracyRadius;
+  }
+
+  // TODO: 12/07/2018 canceling zoom and tilt
```
danesfeder: This is okay to go into `master` as is?

LukasPaczos: Whoops, definitely not. Good catch.

LukasPaczos:
We also need to update the `android-docs` after this changes are released to document the new `LocationLayerPlugin#zoomWhileTracking` and `LocationLayerPlugin#tiltWhileTracking` methods.

/cc @langsmith

PR #584, opened 2018-07-18
Fix for disappearing places picker marker

### LINKED ISSUE #529: Marker icon is drifting away
In the demo app in Place picker the marker icon is drifting down from the center of the screen. This happens when I swipe the map up and then quickly after it swipe the map down. After the maker is no longer in the center, it moves down every time I move the map and as a result disappears from the screen. Only the shadow remains on its place.
device-2018-05-22-110422
device-2018-05-22-110443
device-2018-05-22-110510

langsmith:
@tobrun @LukasPaczos 

Resolves #529 by fixing the floating marker in the place picker functionality of the places plugin.  In my opinion, this bug renders the picker essentially unusable.

If you share my slight concern that`mapView.getY() / 2` isn't the most performant fix, given that it happens every time that `onCameraIdle()` fires, I'm happy to hear other ideas for fixing.

ezgif com-optimize

### DIFF
--- a/plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/picker/ui/PlacePickerActivity.java
+++ b/plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/picker/ui/PlacePickerActivity.java
@@ -143,7 +143,7 @@ public class PlacePickerActivity extends AppCompatActivity implements OnMapReady
   @Override
   public void onCameraIdle() {
     Timber.v("Map camera is now idling.");
-    markerImage.animate().translationY(Math.abs(mapView.getY() / 2))
+    markerImage.animate().translationY(0)
       .setInterpolator(new OvershootInterpolator()).setDuration(250).start();
     bottomSheet.setPlaceDetails(null);
     makeReverseGeocodingSearch();


plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/picker/ui/PlacePickerActivity.java
```diff
@@ -143,7 +143,7 @@ public void onCameraMoveStarted(int reason) {
   @Override
   public void onCameraIdle() {
     Timber.v("Map camera is now idling.");
-    markerImage.animate().yBy(Math.abs(markerImage.getTranslationY()))
+    markerImage.animate().translationY(Math.abs(mapView.getY() / 2))
```
LukasPaczos: Actually, as the `markerImage` has view constraints to its parent, just `markerImage.animate().translationY(0)` should be enough here.

langsmith: Yea, that works. Fixed

ezgif com-optimize

langsmith:
Any chance we can do a new release of the Places Plugin in the next couple of weeks @LukasPaczos @zugaldia ? A release hasn't gone out since this pr was merged into `master`, so in my opinion, the place picker functionality continues to be broken/useless for devs who are using the latest stable version. They'll still see #529 in their projects. Or are we fine with advising folks to use `0.6.0-SNAPSHOT` for now? 

cc @riastrad in case you get any future inbound questions about this.

bradjonesca:
@zugaldia could you explain or direct me to any information about the "frappe" release? I had not heard of it before now, thank-you

zugaldia:
@bradjonesca my bad, I should have provided a better explanation :) `frappe` is the codename for the upcoming Maps SDK release, which on Android is 6.5.0. We just released 6.5.0-alpha.2 and we expect a final release in the next few weeks (we name our Maps SDK release after drinks, you can see previous names in the `release-*` branch names in http://github.com/mapbox/mapbox-gl-native).

PR #614, opened 2018-08-04
Fix compass for devices without gyroscope

### LINKED ISSUE #613: Compass not working on devices without gyroscope
I've been testing my app on my Moto 3G (2015) API 25 and noticed, that compass tracking wasn't working, it just stands still. Similar problem I've encountered with Sky Map application, I should go to the settings of the app and disable gyroscope. 

Then I've tested plugin-locationlayer on 2 different emulators Nexus 5X API 27 and on Nexus S API 22. It works on Nexus 5X, but not working on Nexus S. After listing all available sensors on all of this three devices appears that on all devices CompassManager falls back to Sensor.TYPE_ORIENTATION, but it works only on device with gyroscope. 

So I propose in case of fallback add check does device have gyroscope, and if it has use Sensor.TYPE_ORIENTATION, otherwise fallback further to combination of accelerometer and magnetic field sensor. 

I've already tested this and can make pull request if it's appropriate solution to futher investiagation.

niihelium:
Fixes #613 

As I've mentioned in issue if there is no gyroscope sensor on the device we can't use TYPE_ORIENTATION software sensor, also usage of TYPE_GEOMAGNETIC_ROTATION_VECTOR is not possible due to min API requirement. So here is proposed solution with usage of accelerometer and magnetic sensor with help of low pass filter for smoothing.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CompassManager.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CompassManager.java
@@ -38,6 +38,10 @@ class CompassManager implements SensorEventListener {
   // Not all devices have a compassSensor
   @Nullable
   private Sensor compassSensor;
+  @Nullable
+  private Sensor gravitySensor;
+  @Nullable
+  private Sensor magneticFieldSensor;
 
   private float[] truncatedRotationVectorValue = new float[4];
   private float[] rotationMatrix = new float[9];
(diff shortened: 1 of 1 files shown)


niihelium:
@danesfeder can you, please, review this one, or ask someone and make comments.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/CompassManager.java
```diff
@@ -81,7 +100,12 @@ void removeCompassListener(@NonNull CompassListener compassListener) {
   void onStart() {
     if (isSensorAvailable()) {
       // Does nothing if the sensors already registered.
-      sensorManager.registerListener(this, compassSensor, SENSOR_DELAY_MICROS);
+      if (compassSensor != null) {
+        sensorManager.registerListener(this, compassSensor, SENSOR_DELAY_MICROS);
+      }else {
```
danesfeder: Code formatting > `} else {`

PR #626, opened 2018-08-27
Add getter and setter for CompassEngine interface

### LINKED ISSUE #617: LocationLayerPlugin -- Make CompassManager extensible
Allow developers to create their own `CompassManager` to provide their own bearing and heading info to the plugin.  

For example, would allow developers to alternate between sensor data and snapped location data from the navigation SDK.

cc @cmahopper @hurrba @LindaRosa

danesfeder:
Closes #617 

Allows custom `CompassEngine` for generating bearing values provided to the `LocationLayerPlugin`

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
@@ -73,7 +73,7 @@ public final class LocationLayerPlugin implements LifecycleObserver {
   private final MapView mapView;
   private LocationLayerOptions options;
   private LocationEngine locationEngine;
-  private CompassManager compassManager;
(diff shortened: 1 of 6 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerPlugin.java
```diff
@@ -519,6 +519,28 @@ public LocationEngine getLocationEngine() {
...
   }
 
+  /**
+   * Sets the compass engine used to provide compass heading values.
+   *
+   * @param compassEngine to be used
+   * @since 0.8.0
+   */
+  public void setCompassEngine(@NonNull CompassEngine compassEngine) {
+    this.compassEngine = compassEngine;
```
LukasPaczos: I think we should remove the listener attached to the current `CompassEngine` before swapping it.

PR #641, opened 2018-09-03
[LLP] prevent from showing the location layer on every location update

Guardiola31337:
- Prevents from showing the location layer on every location update 

Aims to fix the performance 📉 shown in https://github.com/mapbox/mapbox-navigation-android/pull/1252

I was able to reproduce in `master` 👀 

llp_jumpy

The idea is to call `locationLayer.show()` method only once (when the first valid fix is received).

Apparently, calling `show` > `setRenderMode` > `LocationLayer#styleForeground` > `generateBitmap` multiple times - especially when the `RenderMode` is `GPS` in which `generateBitmap` is called 4 times - is quite heavy and has a noticeable impact on low-end devices. This was tested on a Nexus 5 running Android 6.0

cc @danesfeder

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayer.java
@@ -168,6 +168,10 @@ final class LocationLayer implements PluginAnimator.OnLayerAnimationsValuesChang
     }
   }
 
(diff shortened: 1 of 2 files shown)


Guardiola31337:
@LukasPaczos 
> I think it'd be cleaner to expose and use `LocationLayer#isHidden` flag instead of introducing a new one that tries to accomplish the same thing

Yeah, that's a great idea! Thanks for the quick review here. This is ready for a second round of 👀

PR #652, opened 2018-09-06
Add MarkerView plugin

tobrun:
Upstream in the Mapbox Maps SDK for Android, we have deprecated MarkerView for [almost a year](https://github.com/mapbox/mapbox-gl-native/pull/9782). While this API can unblock certain use-cases,  it isn't ideal and has no place in the main SDK. We plan to remove support of this API in the next major version (v7.0.0). To not block users still using this deprecated API, we are moving this feature inside a plugin instead.

ezgif com-video-to-gif 39


plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
```diff
@@ -0,0 +1,67 @@
...
+  private final MapboxMap mapboxMap;
+  private final List<MarkerView> markers = new ArrayList<>();
+  private boolean initalised;
+
+  public MarkerViewManager(MapView mapView, MapboxMap mapboxMap) {
+    this.mapView = mapView;
+    this.mapboxMap = mapboxMap;
+  }
+
+  public void onDestroy() {
```
LukasPaczos: This call is missing in the example. This method should probably be limited to the main thread as well.

plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
```diff
@@ -0,0 +1,67 @@
...
+    mapView.removeOnMapChangedListener(this);
+    initalised = false;
+  }
+
+  @UiThread
+  public void addMarker(@NonNull MarkerView markerView) {
+    if (markers.contains(markerView)) {
```
LukasPaczos: We should check here for `MapView#isDestroyed`, not to hook the listener into a destroyed map.

PR #654, opened 2018-09-07
Update localization plugin to support expressions

tobrun:
This PR updates our localization plugin to support expressions  (also backwards compatible with our old token syntax setup). Atm it requires a beta.1 build of the Mapbox Maps SDK for Android, as it requires https://github.com/mapbox/mapbox-gl-native/pull/12769 and https://github.com/mapbox/mapbox-gl-native/pull/12768 to be fixed upstream. We will have to hold off merging and releasing this plugin until 6.5.0 final has been released, eta 1 week. 

One issue I had with migration is that some step expression returned from core contain an invalid representation. On [this](https://github.com/mapbox/mapbox-plugins-android/compare/tvn-upgrade-local-to-expressions?expand=1#diff-1bf159640ccc4e006c457ee997362bf9R153) line I'm fixing this up by introducing an additional 0 stop output based on the current locale. This fix will not break if the issue is fixed upstream though would love to remove it at some point.

### DIFF
--- a/plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
+++ b/plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
@@ -1,20 +1,23 @@
 package com.mapbox.mapboxsdk.plugins.localization;
 
 import android.support.annotation.NonNull;
-
 import com.mapbox.mapboxsdk.camera.CameraUpdateFactory;
 import com.mapbox.mapboxsdk.geometry.LatLngBounds;
 import com.mapbox.mapboxsdk.maps.MapView;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
 import com.mapbox.mapboxsdk.plugins.localization.MapLocale.Languages;
+import com.mapbox.mapboxsdk.style.expressions.Expression;
 import com.mapbox.mapboxsdk.style.layers.Layer;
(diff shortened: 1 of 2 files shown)


plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
```diff
@@ -104,20 +122,40 @@ public void setMapLanguage(@NonNull MapLocale mapLocale) {
...
+    }
+  }
+
+  private void convertExpression(@NonNull MapLocale mapLocale, Layer layer, PropertyValue<?> textFieldProperty) {
+    Expression textFieldExpression = textFieldProperty.getExpression();
+    if (textFieldExpression != null) {
+      String text = textFieldExpression.toString().replaceAll(EXPRESSION_REGEX, mapLocale.getMapLanguage());
+      if (text.startsWith("[\"step") && textFieldExpression.toArray().length % 2 == 0) {
+        // got an invalid step expression from core, we need to add an additional name_x into step
+        text = text.replaceAll(STEP_REGEX, String.format(STEP_TEMPLATE, mapLocale.getMapLanguage()));
```
1ec5: Yep, we’re doing this on iOS/macOS too:

https://github.com/mapbox/mapbox-gl-native/blob/79bf0e8af6bf9ec829a352d56b8e70ccc8f4fa41/platform/darwin/src/NSExpression%2BMGLAdditions.mm#L796-L799

PR #676, opened 2018-09-14
Introduce abstract Annotation and AnnotationManager 

tobrun:
This PR pulls some shared logic into a base class, this should improve maintainability of the generated code.


LukasPaczos:
I like this changes! Also, looking through https://github.com/mapbox/mapbox-plugins-android/pull/685 where we need to add a lot of duplicating code to handle a different type of click - do you see an option here to generify the layer initialization and click resolvers as well? It seems like the only missing pieces in order to achieve that are layer/source IDs which could be provided with an abstract method.

PR #688, opened 2018-09-18
Add generated activity sanity test using ejs

tobrun:
This adds generated activity tests using ejs and removes robo tests from ci runs


app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/activity.junit.ejs
```diff
@@ -0,0 +1,31 @@
+<%
+  const activity = locals[0];
+  const subPackage = locals[1];
+-%>
+// This file is generated. Edit android/platform/scripts/generate-test-code.js, then run `make generate-test-android`.
```
LukasPaczos: This note should be updated.

PR #700, opened 2018-09-25
Add builder classes

### LINKED ISSUE #690: Add option classes
Next to performance improvements shown in https://github.com/mapbox/mapbox-plugins-android/issues/679 and https://github.com/mapbox/mapbox-plugins-android/issues/680. We should expose builder classes. This allows to configure an annotation but not to dispatch any updates until the annotation is ready to be shown. This allows to minimise source updates as much as possible.

### LINKED ISSUE #696: Remove self update mechanism
Every time we update an annotation property we update the source. We should move the responsibility of updating to the end developer to allow to change multiple properties performantly. For example changing the icon, text and location of a symbol will result in 3 updates while these can be consolidated into one.  API wise this will result in making our annotations classes into POJO classes and exposing an `AnnotationManager#update` method to perform the actual update of the underlying source.

tobrun:
Closes #690
Update: closes #696


tobrun:
Merged in https://github.com/mapbox/mapbox-plugins-android/pull/702 which removes self updates from the model classes.

tobrun:
I'm going to make the options classes more generic so we can consolidate more into the abstract AnnotationManager. Idea is to have all CRUD operations generically defined in that class.

PR #706, opened 2018-09-28
Add animation duration multiplier configuration

tobrun:
This PR adds the ability to configure the animation duration multiplier used.

### DIFF
--- a/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
+++ b/plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
@@ -75,6 +75,11 @@ public class LocationLayerOptions implements Parcelable {
    */
   private static final long STALE_STATE_DELAY_MS = 30000;
 
+  /**
+   * Default animation duration multiplier
+   */
+  private static final float TRACKING_ANIMATION_DURATION_MULTIPLIER_DEFAULT = 1.1f;
+
   private float accuracyAlpha;
   private int accuracyColor;
   private int backgroundDrawableStale;
@@ -105,6 +110,7 @@ public class LocationLayerOptions implements Parcelable {
   private float trackingInitialMoveThreshold;
   private float trackingMultiFingerMoveThreshold;
   private String layerBelow;
+  private float trackingAnimationDurationMultiplier;
 
   public LocationLayerOptions(
     float accuracyAlpha,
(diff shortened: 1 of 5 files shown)


plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
```diff
@@ -1534,6 +1564,17 @@ private Builder(LocationLayerOptions source) {
       return this;
     }
 
+    /**
+     * Sets the animation duration multiplier.
+     *
+     * @param animationDurationMultiplier the animation duration multiplier
+     * @since 0.9.0
+     */
+    public LocationLayerOptions.Builder animationDurationMultiplier(float animationDurationMultiplier) {
```
LukasPaczos: Could we indicate that this is a tracking animation duration? Either in the method name, javadoc or both works.

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
```diff
@@ -748,6 +769,7 @@ public String toString() {
       + "trackingInitialMoveThreshold=" + trackingInitialMoveThreshold + ", "
       + "trackingMultiFingerMoveThreshold=" + trackingMultiFingerMoveThreshold + ", "
       + "layerBelow=" + layerBelow
+      + "animationDurationMultiplier=" +animationDurationMultiplier
```
LukasPaczos: NIT whitespace missing

plugin-locationlayer/src/main/java/com/mapbox/mapboxsdk/plugins/locationlayer/LocationLayerOptions.java
tobrun: Yes, went with updating all occurrences of the name by pre-preding tracking to it so it conforms to other tracking mode configurations. Thanks for the quick 👀.

LukasPaczos:
Changing the multiplier at this point requires recreating the plugin. If that should be the case, javadoc entry would be great, otherwise, we should reload the multiplier in `PluginAnimatorCoordinator` whenever new options are provided.

tobrun:
Thanks for catching that, in https://github.com/mapbox/mapbox-plugins-android/pull/706/commits/6b0a17642e87d574ecaca3368f83b23bc660a725 I replaced the constructor configuration with a setter. That felt cleaner as recreating the object. Ready for review :)

PR #722, opened 2018-10-12
use PointF instead of Float[]

tobrun:
Similar to https://github.com/mapbox/mapbox-plugins-android/pull/721, this PR optimises the API to take in a PointF instead of a Float[] for our offset API. Reasoning behind this is that a PointF feels more natural to work with for an Android developer (see similar feature request upstream in https://github.com/mapbox/mapbox-gl-native/issues/11499).

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
@@ -2,6 +2,7 @@
 
 package com.mapbox.mapboxsdk.plugins.annotation;
 
+import android.graphics.PointF;
 import android.support.annotation.UiThread;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
@@ -78,6 +79,7 @@ public class Symbol extends Annotation {
   //}
 
   // Property accessors
+
   /**
    * Get the IconSize property
    *
@@ -144,15 +146,15 @@ public class Symbol extends Annotation {
   /**
    * Get the IconOffset property
    *
-   * @return property wrapper value around Float[]
+   * @return PointF value for Float[]
    */
-  public Float[] getIconOffset() {
+  public PointF getIconOffset() {
     JsonArray jsonArray = jsonObject.getAsJsonArray("icon-offset");
     Float[] value = new Float[jsonArray.size()];
     for (int i = 0; i < jsonArray.size(); i++) {
       value[i] = jsonArray.get(i).getAsFloat();
     }
-    return value;
+    return new PointF(value[0], value[1]);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
```diff
@@ -144,25 +146,26 @@ public void setIconRotate(Float value) {
...
    * Set the IconOffset property.
    * <p>
    * To update the symbol on the map use {@link SymbolManager#update(Annotation)}.
    * <p>
-   * @param value constant property value for Float[]
+   * @param pointF value for Float[]
    */
-  public void setIconOffset(Float[] value) {
+  public void setIconOffset(PointF pointF) {
+    Float[] value = new Float[]{pointF.x, pointF.y};
```
LukasPaczos: This seems like an unnecessary object allocation, we could insert `pointF.x` and `pointF.y` to the JSON array manually as we know there will always be 2 values, no need to iterate through them.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
```diff
@@ -144,25 +146,26 @@ public void setIconRotate(Float value) {
   /**
    * Get the IconOffset property
    *
-   * @return property wrapper value around Float[]
+   * @return PointF value for Float[]
    */
-  public Float[] getIconOffset() {
+  public PointF getIconOffset() {
     JsonArray jsonArray = jsonObject.getAsJsonArray("icon-offset");
     Float[] value = new Float[jsonArray.size()];
```
LukasPaczos: This also seems like an unnecessary object allocation, we could create `PointF` immediately.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
```diff
@@ -391,25 +394,26 @@ public void setTextTransform(@Property.TEXT_TRANSFORM String value) {
   /**
    * Get the TextOffset property
    *
-   * @return property wrapper value around Float[]
+   * @return PointF value for Float[]
    */
-  public Float[] getTextOffset() {
+  public PointF getTextOffset() {
     JsonArray jsonArray = jsonObject.getAsJsonArray("text-offset");
     Float[] value = new Float[jsonArray.size()];
```
LukasPaczos: Same as above.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Symbol.java
```diff
@@ -391,25 +394,26 @@ public void setTextTransform(@Property.TEXT_TRANSFORM String value) {
...
    * Set the TextOffset property.
    * <p>
    * To update the symbol on the map use {@link SymbolManager#update(Annotation)}.
    * <p>
-   * @param value constant property value for Float[]
+   * @param pointF value for Float[]
    */
-  public void setTextOffset(Float[] value) {
+  public void setTextOffset(PointF pointF) {
+    Float[] value = new Float[]{pointF.x, pointF.y};
```
LukasPaczos: Same as above.

tobrun:
Thank you for the review @LukasPaczos , addressed our comments in https://github.com/mapbox/mapbox-plugins-android/pull/722/commits/4baffafe5c1637fc10035e959fa099cc82a7be1a

Issue #724, opened 2018-10-16
OfflineDownloadChangeListener not working


arnekaiser:
None of the callback methods are called anymore. The problem appears also in the test app of this repo (see OfflineRegionDetailActivity).

Tested on OnePlus 5 (Android 8.1) and emulator (Android 9).

tobrun:
Fixed with https://github.com/mapbox/mapbox-plugins-android/commit/452f25e830fc904aa7fd8508d3d4daacc1750d5c

PR #725, opened 2018-10-16
Fixed: OfflineDownloadChangeListener not working #724

arnekaiser:
The broadcast receiver was not initialized.


tobrun:
Thank you for catching this and providing a fix!

PR #732, opened 2018-10-25
Plugin README link updates

langsmith:
README links to test app activities were broken because we switched the examples to Kotlin. This pr fixes the links.


langsmith:
That did the trick

Issue #703, opened 2018-09-26
Draggable symbols


tobrun:
One highly requested feature on the annotation side upstream in the maps sdk was draggable markers. Let's implement this on symbols, see reference implementation [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDKTestApp/src/main/java/com/mapbox/mapboxsdk/testapp/activity/style/DraggableMarkerActivity.kt).

PR #753, opened 2018-11-05
Draggable annotations

LukasPaczos:
Closes https://github.com/mapbox/mapbox-plugins-android/issues/703.

ezgif com-video-to-gif 14

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
@@ -5,6 +5,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.UiThread;
 import android.support.v4.util.LongSparseArray;
+
 import com.mapbox.geojson.Feature;
 import com.mapbox.geojson.FeatureCollection;
 import com.mapbox.mapboxsdk.geometry.LatLng;
@@ -21,12 +22,14 @@ import java.util.List;
  *
  * @param <T> type of annotation
  * @param <S> type of options for building the annotation, depends on generic T
+ * @param <D> type of annotation drag listener, depends on generic T
  * @param <U> type of annotation click listener, depends on generic T
  * @param <V> type of annotation long click listener, depends on generic T
  */
-public abstract class AnnotationManager<
+abstract class AnnotationManager<
   T extends Annotation,
   S extends Options<T>,
+  D extends OnAnnotationDragListener<T>,
   U extends OnAnnotationClickListener<T>,
   V extends OnAnnotationLongClickListener<T>> {
 
@@ -34,6 +37,8 @@ public abstract class AnnotationManager<
   protected final LongSparseArray<T> annotations = new LongSparseArray<>();
   protected final List<Feature> features = new ArrayList<>();
 
+  private final DraggableAnnotationController<T, D> draggableAnnotationController;
+  private final List<D> dragListeners = new ArrayList<>();
   private final List<U> clickListeners = new ArrayList<>();
   private final List<V> longClickListeners = new ArrayList<>();
   protected long currentId;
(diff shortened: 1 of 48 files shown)


LukasPaczos:
This one is ready for a review.

ezgif com-video-to-gif 15

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
```diff
@@ -265,16 +295,26 @@ public void onMapLongClick(@NonNull LatLng point) {
...
+  }
+
+  @Nullable
+  T queryMapForFeatures(@NonNull PointF point) {
+    List<Feature> features = mapboxMap.queryRenderedFeatures(point, getAnnotationLayerId());
+    if (!features.isEmpty()) {
+      long id = features.get(0).getProperty(getAnnotationIdKey()).getAsLong();
+      return annotations.get(id);
     }
+    return null;
```
tobrun: there seems to be some shared code between these 2 methods.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Circle.java
```diff
@@ -35,12 +42,23 @@
...
   }
 
+  /**
+   * Get the LatLng of the circle, which represents the location of the circle on the map
+   *
+   * @return the location of the circle
+   */
+  @NonNull
+  public LatLng getLatLng() {
+    Point point = (Point) geometry;
```
tobrun: not related to this PR, but something that comes to mind now, do we need type geometry with generics? (avoids casting in this case).

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Fill.java
```diff
@@ -49,6 +56,29 @@ public void setLatLngs(List<List<LatLng>> latLngs) {
...
+      for (List<Point> innerPoints : coordinates) {
+        List<LatLng> innerList = new ArrayList<>();
+        for (Point point : innerPoints) {
+          innerList.add(new LatLng(point.latitude(), point.longitude()));
+        }
+        latLngs.add(innerList);
+      }
+      return latLngs;
+    }
+    return null;
```
tobrun: do we want nullability for this? Returning an empty list would suffice and would make the method `@NonNull`

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Fill.java
```diff
@@ -148,4 +187,33 @@ public String getFillPattern() {
...
+          }
+          innerPoints.add(Point.fromLngLat(latLng.getLongitude(), latLng.getLatitude()));
+        }
+        resultingPoints.add(innerPoints);
+      }
+
+      return Polygon.fromLngLats(resultingPoints);
+    }
+
+    return null;
```
tobrun: do we want nullability for this? Returning an empty list would suffice and would make the method `@NonNull`

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Fill.java
LukasPaczos: Good point

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Fill.java
LukasPaczos: In this case, we are returning `null` to indicate that the resulting geometry is invalid and we should not set it, leaving the old one being displayed

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
LukasPaczos: yup, going to fix this up

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/Circle.java
LukasPaczos: we can possibly do it to clean up those casts :+1:

PR #774, opened 2018-11-19
Add expression for dds supported properties on managers

### LINKED ISSUE #736: Expression support
The annotation plugin attempts to provide a simplified layer API that is more object driven. In some cases you would still like to use expressions in combination with the annotation plugin API. We could expose expression setters for DDS properties on the manager classes that would override the fixed value setup from the annotation class.

tobrun:
Closes #736. PR exposes basic setters and getters for expression on the manager level. They override behaviour defined in the underlying annotation. This enables users to combine per annotation syntax with using expressions on the manager class. Basic testing for boolean, number and string expression values were added. Leaving out format, array,.. out for now.


tobrun:
@LukasPaczos re overriding behaviour, to me, that code snippet is expected behaviour. I don't mind expanding adding additional logic to the handle that kind of case. The idea is to add some kind of blacklist? Once you set an expression you can't override it with an annotation implementation?

LukasPaczos:
It might not feel intuitive for a user not familiar with how runtime/dd styling works under the hood that setting a property on a single annotation changes the behavior for all annotations within a manager. To help this, we can add the blacklist you mentioned - if an expression is set, ignore annotation property and print a warning if set, and add a `removeExpression` method to cycle back to the default behavior.

LukasPaczos:
We can, later on, expand this by exposing expression setters on annotations themselves and dynamically generate the layer property expression.

PR #776, opened 2018-11-20
Add symbol animation example

tobrun:
PR adds a simple animation example to look into and close https://github.com/mapbox/mapbox-plugins-android/issues/770.

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
@@ -1,5 +1,8 @@
 package com.mapbox.mapboxsdk.plugins.testapp.activity.annotation;
 
+import android.animation.Animator;
+import android.animation.AnimatorListenerAdapter;
+import android.animation.ValueAnimator;
 import android.graphics.Color;
 import android.graphics.PointF;
 import android.os.Bundle;
@@ -7,10 +10,12 @@ import android.support.v7.app.AppCompatActivity;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.animation.LinearInterpolator;
 import android.widget.TextView;
 import android.widget.Toast;
 
 import com.mapbox.geojson.FeatureCollection;
+import com.mapbox.geojson.Point;
 import com.mapbox.mapboxsdk.camera.CameraUpdateFactory;
 import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.maps.MapView;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
```diff
@@ -161,35 +166,86 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
     return true;
   }
 
+  private void resetSymbol() {
+    symbol.setIconRotate(0.0f);
+    symbol.setGeometry(Point.fromLngLat(6.687337, 0.381457));
+    symbolManager.update(symbol);
+  }
+
+
```
LukasPaczos: NIT: new line

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
```diff
@@ -161,35 +166,86 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+  private void easeSymbol(com.mapbox.mapboxsdk.plugins.annotation.Symbol symbol, final LatLng location, final float rotation) {
```
LukasPaczos: NIT: package name

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
```diff
@@ -161,35 +166,86 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+    final LatLng originalPosition = symbol.getLatLng();
+    final float originalRotation = symbol.getIconRotate();
+    final boolean changeLocation = originalPosition.distanceTo(location) > 0;
+    final boolean changeRotation = originalRotation != rotation;
+    if (!changeLocation && !changeRotation) {
+      return;
+    }
+
+    ValueAnimator moveSymbol = ValueAnimator.ofFloat(0, 1).setDuration(5000);
```
LukasPaczos: Reference to this animator should be kept and animation canceled when the new one starts (with a menu option click) or when the activity is stopped/finished. Otherwise, we are leaking and crashing. We can also harden this by refusing source updates after `AnnotationManager#onDestroy` call and logging an error (similarly to `NativeMapView` impl, this would prevent the crash but not the leak).

tobrun: I went with cancelling the animator before we stop rendering the map.

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/SymbolActivity.java
```diff
@@ -46,6 +45,7 @@
   private static final String MAKI_ICON_CIRCLE = "circle-15";
 
   private final Random random = new Random();
+  private final List<ValueAnimator>animators  = new ArrayList<>();
```
LukasPaczos: NIT: formatting

PR #778, opened 2018-11-22
Support for streets v8 sources

LukasPaczos:
This PR adds support for `streets.v8` based tile sources and fixes issues when switching from Simplified Chinese or local name to any other locale.

In `streets.v8` sources language fallback logic in case of missing resources should be executed on style level, rather than source level.

Fallback for resources looks as follows:
- For Latin script based local names:
```
Localization -> Local name
```
- For Simplified Chinese:
```
Simplified Chinese -> Local name or English (depending whether local name is Latin script based)
```
- For Traditional Chinese:
```
Traditional Chinese -> Simplified Chinese -> Local name or English (depending whether local name is Latin script based)
```
- Any other:
```
Localization -> English -> Local name
```

/cc @nickidlugash

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/localization/LocalizationActivity.kt
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/localization/LocalizationActivity.kt
@@ -5,14 +5,12 @@ import android.support.v7.app.AppCompatActivity
 import android.view.Menu
 import android.view.MenuItem
 import android.widget.Toast
-
 import com.mapbox.mapboxsdk.maps.MapboxMap
 import com.mapbox.mapboxsdk.maps.OnMapReadyCallback
 import com.mapbox.mapboxsdk.plugins.localization.LocalizationPlugin
 import com.mapbox.mapboxsdk.plugins.localization.MapLocale
 import com.mapbox.mapboxsdk.plugins.testapp.R
 import com.mapbox.mapboxsdk.plugins.testapp.Utils
-
 import kotlinx.android.synthetic.main.activity_localization.*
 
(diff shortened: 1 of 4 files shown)


snkashis:
Is there any ETA available for streets v9,v10,v11 support?
https://github.com/mapbox/mapbox-gl-native/blob/46ddb7c0c1c348696e2e2559355e1910a7abb616/platform/android/MapboxGLAndroidSDK/src/main/res/values/strings.xml#L22

LukasPaczos:
@snkashis `streets-v11` style and all styles based on the `streets-v8` tilesets are supported by the Localization Plugin `v0.6.0` (which is based on the Maps SDK `v6.8.0`) and the Localization Plugin `v0.7.0` (which is based on the Maps SDK `v7.0.0`). Also noting, that we've changed the artifacts naming, see [the README](https://github.com/mapbox/mapbox-plugins-android#maps-sdk-compatibility), so the plugin `v0.7.0` is found under [`mapbox-android-plugin-localization-v7:0.7.0` artifact name](https://github.com/mapbox/mapbox-plugins-android/tree/master/plugin-localization#getting-started).

snkashis:
@LukasPaczos if so, why isnt v11 in  `SUPPORTED_SOURCES`?
https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java#L50

LukasPaczos:
@snkashis `SUPPORTED_SOURCES` indicates supported vector tile sources.
`streets-v11` style is based on the [`streets-v8` tileset](https://docs.mapbox.com/vector-tiles/mapbox-streets-v8/), while `streets-v10`, `streets-v9` and other popular styles are based on the [`streets-v7` tileset](https://docs.mapbox.com/vector-tiles/mapbox-streets-v7/).

PR #789, opened 2018-12-15
Maps SDK 7.0.0 and required breaking changes refactoring

langsmith:
This pr bumps the plugins' Maps SDK dependency to `7.0.0-beta.1` in preparation for the stable `7.0.0` release. This pr also refactors all of the plugins to fix breaking changes.  

https://github.com/mapbox/mapbox-android-demo/pull/917 is related.

After commenting out the test app activities which are related to the `LocationLayerPlugin`, I got the test app running again. All of the non-`LocationLayerPlugin` activities are running and behaving fine for me. I encourage you to switch to this branch, comment out `LocationLayerPlugin` `.kt` files, and check out the app on your own device.

@tobrun & @LukasPaczos :

- In some files, I declare `private Style style;`, set `this.style = mapboxMap.getStyle();` in the initialization of the class, and then use `if (style!=null)` in various places throughout the class before running whatever code. Didn't know if that's preferred or if it's fine for us to just do `mapboxMap.getStyle.XXXX` instead.

- Wasn't sure how to handle the mix of breaking changes and the deprecation of the location layer plugin. How should we proceed with it?


langsmith:
Also, other than the `LocationLayerPlugin` tests, I've checked/fixed the plugins' unit/instrumentation tests

langsmith:
>Let's remove the source code of the LocationLayerPlugin, we shouldn't need to maintain that code any longer. I think we can leave the module and README.md not to break existing links that are out there.

@LukasPaczos , this means that the latest/current version of the plugin ( `0.11.0` ) is the final version, right?    Asking because I'm thinking about whether any additional warning(s) about this need to go on https://www.mapbox.com/android-docs/plugins/overview/location-layer  (cc @colleenmcginnis )

LukasPaczos:
@langsmith correct, `0.11.0` is the final version of the `LocationLayerPlugin`.

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/offline/OfflineRegionDetailActivity.kt
```diff
@@ -139,30 +140,30 @@ class OfflineRegionDetailActivity : AppCompatActivity(), OfflineDownloadChangeLi
...
+                // update textview data
+                offlineRegion?.metadata?.let {
+                    regionName.text = OfflineUtils.convertRegionName(it)
+                }
+                regionStyleUrl.text = definition.styleURL
+                regionLatLngBounds.text = definition.bounds.toString()
+                regionMinZoom.text = definition.minZoom.toString()
+                regionMaxZoom.text = definition.maxZoom.toString()
+                offlineRegion?.getStatus(offlineRegionStatusCallback)
+            }
```
tobrun: just as an FYI, we have a method on MapboxMap to make above a bit more easy:
`MapboxMap#setOfflineRegionDefinition(@NonNull OfflineRegionDefinition definition, @Nullable Style.OnStyleLoaded callback)`

plugin-annotation/src/test/java/com/mapbox/mapboxsdk/plugins/annotation/FillManagerTest.java
```diff
@@ -28,18 +31,106 @@
...
+
+    Expression filter = Expression.literal(false);
+    fillManager.setFilter(filter);
+
+    ArgumentCaptor<MapView.OnWillStartLoadingMapListener> loadingArgumentCaptor = ArgumentCaptor.forClass(MapView.OnWillStartLoadingMapListener.class);
+    verify(mapView).addOnWillStartLoadingMapListener(loadingArgumentCaptor.capture());
+    loadingArgumentCaptor.getValue().onWillStartLoadingMap();
+
+    ArgumentCaptor<Style.OnStyleLoaded> styleLoadedArgumentCaptor = ArgumentCaptor.forClass(Style.OnStyleLoaded.class);
+    verify(mapboxMap).getStyle(styleLoadedArgumentCaptor.capture());
```
tobrun: interesting construct this ArgumentCaptor

PR #844, opened 2019-02-07
Maps SDK 7.1.2 bump

langsmith:
Part of the Maps SDK 7.1.2 release

https://github.com/mapbox/gl-internal/issues/1075


langsmith:
>Could we add the version name to the commit message? just bump doesn't add much context when going through the git history.

✔️ fixed

PR #867, opened 2019-03-07
Update SymbolOptions.java

ben-j69:
I did not find any info on the offset, neither in the object neither in the documentation so I think it is better to add in the code doc what we are expecting from it.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/SymbolOptions.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/SymbolOptions.java
@@ -116,7 +116,7 @@ public class SymbolOptions extends Options<Symbol> {
   /**
    * Set icon-offset to initialise the symbol with.
    *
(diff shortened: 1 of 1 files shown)


tobrun:
Thank you for the contribution @ben-j69, the file you are updating however is generated, see the top level deceleration [here](https://github.com/mapbox/mapbox-plugins-android/blob/eb9c6c671d656b25f062248d2b0dfcc883787c91/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/SymbolOptions.java#L1). You will have to update the template file and the code generation scripts [here](https://github.com/mapbox/mapbox-plugins-android/tree/master/plugin-annotation/scripts). You can run these scripts with the [this](https://github.com/mapbox/mapbox-plugins-android/blob/master/Makefile#L32) make command.

tobrun:
Ticketed out https://github.com/mapbox/mapbox-plugins-android/issues/869 to fix code documentation generation in total. Thank you for wanting to contribute @ben-j69 but closing this as we need a code gen solution for this part of the code.

PR #874, opened 2019-03-12
[test-app] Fixed possible NullPointerException

zhao-gang:
If map style can't be loaded(for example no Internet), annotation
managers won't be initiated. And calling these managers' onDestroy
method would throw a NullPointerException. Fixed it by checking if the
manager is null before calling its onDestroy method.


tobrun:
Thank you for the contribution!

PR #878, opened 2019-03-14
 [localization] Add better support for Chinese languages (fix #877)

### LINKED ISSUE #877: [localization] Add better support for Chinese languages
Different mapbox-streets source versions have different levels/types
of support for Chinese languages. To support Chinese languages, it's
better to map android system Locale to MapLocale dynamically according
to the mapbox-streets source version that is being used.

Android system's Chinese Locales also changed in different android
versions. Add support for new Locales like zh_CN_#Hans (with script
part in Locale).

zhao-gang:
This PR aims to add better support for Chinese languages in localization plugin. With this change the plugin can support more types of Android system Chinese Locales and has better support for mapbox-streets-v8's new name_zh_Hant Traditional Chinese labels.

### DIFF
--- a/plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
+++ b/plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
@@ -231,7 +231,6 @@ public final class LocalizationPlugin {
     for (Source source : style.getSources()) {
       if (sourceIsFromMapbox(source)) {
         boolean isStreetsV8 = sourceIsStreetsV8(source);
-        boolean isStreetsV7 = sourceIsStreetsV7(source);
         for (Layer layer : layers) {
           if (layer instanceof SymbolLayer) {
             PropertyValue<?> textFieldProperty = ((SymbolLayer) layer).getTextField();
@@ -239,6 +238,7 @@ public final class LocalizationPlugin {
               if (isStreetsV8) {
                 convertExpressionV8(mapLocale, layer, textFieldProperty);
               } else {
+                boolean isStreetsV7 = sourceIsStreetsV7(source);
                 convertExpression(mapLocale, layer, textFieldProperty, isStreetsV7);
               }
             }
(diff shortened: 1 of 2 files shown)


plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
```diff
@@ -277,11 +356,6 @@ private void convertExpressionV8(@NonNull MapLocale mapLocale, Layer layer, Prop
 
       String mapLanguage = mapLocale.getMapLanguage();
       if (!mapLanguage.equals(MapLocale.ENGLISH)) {
-        if (mapLanguage.equals(MapLocale.CHINESE)) {
-          // in streets v8 tiles chinese is declared as "name_zh-Hant" instead of "name_zh"
-          mapLanguage = MapLocale.CHINESE_V8;
```
LukasPaczos: This is still needed, otherwise, we'll break the compatibility of `MapLocale.CHINESE` with `streets-v8` sources.

LukasPaczos: This can be ignored if the proposed solution is implemented.

zhao-gang:
@LukasPaczos Hi. I think your suggestion is reasonable and have made the changes you requested. Also rebased this PR to current master.

plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
LukasPaczos: With the current implementation, this is still needed.

plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/LocalizationPlugin.java
```diff
@@ -214,14 +231,15 @@ public void setMapLanguage(@NonNull MapLocale mapLocale) {
     for (Source source : style.getSources()) {
       if (sourceIsFromMapbox(source)) {
         boolean isStreetsV8 = sourceIsStreetsV8(source);
+        boolean isStreetsV7 = sourceIsStreetsV7(source);
```
LukasPaczos: This check can be moved to line 242, otherwise, it's going to be redundant for `streets-v8` sources.

zhao-gang: I have made above two changes(`name_zh` should be changed to `name_zh_Hans` in streets v8, as they both represent Simplified Chinese). I also changed Locale.CHINA's default MapLocale to MapLocale.CHINESE_HANS, since in streets v8 Simplified Chinese should use `name_zh_Hans`.

PR #885, opened 2019-03-18
Notify gesture manager when drag is handled

sebastienrouif:
Going through the DraggableAnnotationController I realized that the boolean returned were sometimes returning true when events were not consumed

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
@@ -24,7 +24,7 @@ final class DraggableAnnotationController<T extends Annotation, D extends OnAnno
   private final int touchAreaMaxX;
   private final int touchAreaMaxY;
 
-  private T draggedAnnotation;
+  @Nullable private T draggedAnnotation;
 
   @SuppressLint("ClickableViewAccessibility")
   DraggableAnnotationController(MapView mapView, MapboxMap mapboxMap) {
@@ -67,9 +67,8 @@ final class DraggableAnnotationController<T extends Annotation, D extends OnAnno
     if (detector.getPointersCount() == 1) {
       T annotation = annotationManager.queryMapForFeatures(detector.getFocalPoint());
       if (annotation != null) {
-        startDragging(annotation);
+        return startDragging(annotation);
       }
-      return annotation != null;
     }
     return false;
   }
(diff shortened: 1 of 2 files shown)


plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
```diff
@@ -116,25 +121,19 @@ void onMoveEnd() {
     stopDragging(draggedAnnotation);
   }
 
-  void startDragging(@Nullable T annotation) {
-    if (annotation != null && annotation.isDraggable()) {
-      if (!annotationManager.getDragListeners().isEmpty()) {
```
LukasPaczos: This prevents from creating unnecessary iterators, it'd be great to keep it in place.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
```diff
@@ -62,11 +63,15 @@ void onSourceUpdated() {
     stopDragging(draggedAnnotation);
   }
 
-  void onMoveBegin(MoveGestureDetector detector) {
+  boolean onMoveBegin(MoveGestureDetector detector) {
     if (detector.getPointersCount() == 1) {
       T annotation = annotationManager.queryMapForFeatures(detector.getFocalPoint());
-      startDragging(annotation);
+      if (annotation != null) {
+        startDragging(annotation);
```
LukasPaczos: I think `startDragging` should return a `boolean` as well because we still want to return `false` if the annotation is not draggable, or we can move that check up.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
```diff
@@ -116,25 +121,19 @@ void onMoveEnd() {
...
-      }
       draggedAnnotation = annotation;
-    } else {
-      draggedAnnotation = null;
     }
   }
 
   void stopDragging(@Nullable T annotation) {
     if (annotation != null) {
-      if (!annotationManager.getDragListeners().isEmpty()) {
```
LukasPaczos: Same as above.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
sebastienrouif: I thought about making `startDragging` `stopDragging` private as they are only used in tests

LukasPaczos:
Closing in favor of https://github.com/mapbox/mapbox-plugins-android/pull/906, thanks a lot @sebastienrouif!

PR #896, opened 2019-03-30
[annotation-plugin] Avoid adding source/layer twice to style (#888)

zhao-gang:


LukasPaczos:
Thanks a lot for the contribution @zhao-gang! This would've been a great stop-gap solution, but fortunately, we were able to get to the root cause https://github.com/mapbox/mapbox-plugins-android/pull/905.

PR #930, opened 2019-04-12
Remove useless imports

ben-j69:


ben-j69:
I made a fork and bigger changes

PR #931, opened 2019-04-12
Delete unnecessary imports

ben-j69:
A lot of imports are useless in the project


LukasPaczos:
Thanks for the contribution @ben-j69! Those unnecessary imports are in place because the code is generated. Maintaining imports filtering based on the type would be inconvenient in our generator and since the compiler is going to optimize them anyway, we are not worried about leaving them in place.

Closing as redundant, but thanks again for looking into it.

PR #940, opened 2019-04-26
Add getGeometry

### LINKED ISSUE #929: Missing AnnotationOptions#getGeometry
Capturing from @langsmith that this is missing, we expose other successors for other properties.

tobrun:
Closes #929

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/CircleOptions.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/CircleOptions.java
@@ -224,6 +224,18 @@ public class CircleOptions extends Options<Circle> {
   }
 
   /**
+   * Get the LatLng of the circle, which represents the location of the circle on the map
+   *
+   * @return the location of the circle in a longitude and latitude pair
+   */
+  public LatLng getLatLng() {
+    if (geometry == null) {
+      return null;
+    }
+    return new LatLng(geometry.latitude(), geometry.longitude());
+  }
+
+  /**
    * Set the geometry of the circle, which represents the location of the circle on the map
    *
    * @param geometry the location of the circle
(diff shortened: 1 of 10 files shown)


langsmith:
@zestyping , this could help with plugin usage in ODK 👍

zestyping:
Thanks for the heads-up, @langsmith!

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/CircleOptions.java
```diff
@@ -223,6 +223,18 @@ public CircleOptions withLatLng(LatLng latLng) {
...
+  /**
+   * Get the LatLng of the circle, which represents the location of the circle on the map
+   *
+   * @return the location of the circle in a longitude and latitude pair
+   */
+  public LatLng getLatLng() {
+    if (geometry == null) {
+      return null;
+    }
+    return new LatLng(geometry.latitude(), geometry.longitude());
```
LukasPaczos: I feel like the `LatLng` value should be cached next to the `geometry` so that frequent invocations do not lead to unnecessary allocations. What do you think @tobrun?

LukasPaczos: Same for all the other getters.

tobrun: this seems like double bookkeeping and that would require updating the LatLng, List<LatLng>, .. when you update the geometry. This kinda defeats the proposed optimization. My idea is to have the geometry api as fast as possible (animation would be impacted) and  expose the LatLng API for ease of use (where speed doesn't matter else use geometry api instead). thoughts?

LukasPaczos: > animation would be impacted

True, after all, I think it's more beneficial to have fast setter rather than getter. There's always a fallback to raw `geometry` to increase getter performance.

PR #941, opened 2019-04-26
Add multiple manager support

### LINKED ISSUE #921: Multiple manager support
Due to the plugin setup and how gl-native works under the hood, we are unable to show lines with and without a line pattern configuration. Once you set a pattern, the non line pattern lines disappear (more context on this in https://github.com/mapbox/mapbox-gl-native/issues/14331).

One solution to this would be to allow to have multiple line manangers. One for the line pattern one, the other for normal lines. To achieve this we need to generate the IDs of the used source and layers dynamically. 

cc @mapbox/maps-android

tobrun:
Closes #921

### DIFF
--- a/plugin-annotation/scripts/annotation_manager.java.ejs
+++ b/plugin-annotation/scripts/annotation_manager.java.ejs
@@ -22,7 +22,6 @@ import com.mapbox.mapboxsdk.style.layers.<%- camelize(type) %>Layer;
 import com.mapbox.mapboxsdk.style.layers.PropertyValue;
 import com.mapbox.mapboxsdk.style.layers.PropertyFactory;
 import com.mapbox.mapboxsdk.style.sources.GeoJsonOptions;
-import com.mapbox.mapboxsdk.style.sources.GeoJsonSource;
 import com.mapbox.mapboxsdk.style.layers.Property;
 
 import java.util.ArrayList;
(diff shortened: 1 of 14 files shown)


plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -206,8 +207,18 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
...
+  }
+
+  /**
+   * Get the source id of the annotation layer.
+   *
+   * @return the source id
+   */
+  @Override
+  String getSourceId() {
+    return ID_GEOJSON_SOURCE + id;
```
tobrun: Should use String.format instead

plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -206,8 +207,18 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
    * @return the layer id
    */
   @Override
-  String getAnnotationLayerId() {
-    return ID_GEOJSON_LAYER;
+  String getLayerId() {
+    return ID_GEOJSON_LAYER + id;
```
tobrun: Should use String.format instead

tobrun:
Consolidated everything inside concrete instances of CoreElementProvider. Ready for another round of :eyes: .

tobrun:
> A tiny bit more could've been moved to an abstract element provider, but generally LGTM +1

In general yes, good to avoid duplication but since it's generated from a template file, updating code requires updating that one template file in a similar way as you would do with that abstract class. Developer flow wise would kinda be the same.

Issue #854, opened 2019-02-28
Annotations Plugin remove all


w0nderb0y:
I was looking through the code of the SymbolManager but couldn't find a method that clears all symbols. Is there such a thing? Or do I really need to save all my symbols and remove them seperately?

LukasPaczos:
Hey @w0nderb0y! Unfortunately, there is no one, convenient method of removing all of the annotations at once yet. What you can do for now, is iterate through all of the indexes and delete all returned symbols in bundle:
```
List<Symbol> symbols = new ArrayList<>();
LongSparseArray<Symbol> symbolArray = symbolManager.getAnnotations();
for (int i = 0; i < symbolArray.size(); i++) {
  symbols.add(symbolArray.get(i));
}
symbolManager.delete(symbols);
```

w0nderb0y:
hey @LukasPaczos,

Thanks for the very swift reply! 

Let me know once it is implemented. Have a good day

androideveloper:
I was looking for the same method. I will use mapbox.clear() until there is another option to remove them all

androideveloper:
> Hey @w0nderb0y! Unfortunately, there is no one, convenient method of removing all of the annotations at once yet. What you can do for now, is iterate through all of the indexes and delete all returned symbols in bundle:
> 
> ```
> List<Symbol> symbols = new ArrayList<>();
> LongSparseArray<Symbol> symbolArray = symbolManager.getAnnotations();
> for (int i = 0; i < symbolArray.size(); i++) {
>   symbols.add(symbolArray.get(i));
> }
> symbolManager.delete(symbols);
> ```

`symbolArray.get(i)` is searching by key in LongSparseArray. For getting by index `valueAt(i)` should be used.

LukasPaczos:
Correct, thanks @androideveloper.

percula:
For anyone getting this error:

```
2019-04-29 15:37:53.307 A/DEBUG: signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x30
2019-04-29 15:37:53.307 A/DEBUG: Cause: null pointer dereference
2019-04-29 15:37:53.307 A/DEBUG:     x0  00000075ad517a60  x1  00000075ad822cc0  x2  0000000000000040  x3  00000076022c488a
[...]

PR #944, opened 2019-04-30
Harden against crashing when deleting an empty list

tobrun:
Refs https://github.com/mapbox/mapbox-plugins-android/issues/854#issuecomment-487719163, additionally annotates methods from AnnotationManager with null-ability.

cc @percula

### DIFF
--- a/app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/annotation/CircleManagerTest.java
+++ b/app/src/androidTest/java/com/mapbox/mapboxsdk/plugins/annotation/CircleManagerTest.java
@@ -3,6 +3,7 @@
 package com.mapbox.mapboxsdk.plugins.annotation;
 
 import android.support.test.runner.AndroidJUnit4;
+import com.mapbox.geojson.Point;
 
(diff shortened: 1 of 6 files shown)


LukasPaczos:
Capturing from https://github.com/mapbox/mapbox-plugins-android/issues/854#issuecomment-487719163:
> It is caused by calling LineManager.delete(lines) with an empty list.

Is this for sure the cause of the crash? I wasn't able to reproduce with the introduced regression tests after commenting out the `AnnotationManager` changes.

I think it's good to skip the source update if the delete/create list is empty, but it doesn't seem to be related to the crash from the OP.

percula:
I should note it only occurred when updating with an empty list while navigating to a new fragment. I double checked that it was callled from the main thread, but honestly am not sure why it only occurs in that circumstance.

tobrun:
@LukasPaczos hmm yeah, removed the regression test, should be good to merge now. @percula you are calling updates when the underlying style (and thus sources/layers) have been destroyed.

LukasPaczos:
@tobrun could you still include a test to ensure that we are not modifying the annotations array when adding/deleting and options are empty?

PR #946, opened 2019-05-08
[test-app] Update BulkSymbolActivity

zhao-gang:
Don't let Spinner calling OnItemSelectedListener() on activity
start. So LoadLocationTask won't be executed twice. Also let
ProgressDialog be cancelable, as sometimes the ProgressDialog won't
disappear and user can't navigate back and is stuck in this activity.


tobrun:
Thank you for the contribution @zhao-gang !

PR #955, opened 2019-05-15
Add Scale bar plugin

### LINKED ISSUE #11: Scale Widget Plugin
A scale widget plugin allows to place a scalebar on a Mapbox map.

iOS is bringing this feature to their SDK in https://github.com/mapbox/mapbox-gl-native/pull/7631. We could potentially implement this as a plugin instead. This fits in our to decouple our map SDK more into smaller pluggable components.

cc @zugaldia

Chaoba:
Resolves #11 
ezgif com-resize (2)


app/build.gradle
```diff
@@ -113,6 +113,7 @@ dependencies {
   implementation project(':plugin-annotation')
   implementation project(':plugin-markerview')
   implementation project(':ktx-mapbox-maps')
+    implementation project(path: ':plugin-scalebar')
```
tobrun: nit additional spaces

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+import com.mapbox.mapboxsdk.maps.Style
+import com.mapbox.mapboxsdk.plugins.testapp.R
+import com.mapbox.pluginscalebar.ScaleBar
+import kotlinx.android.synthetic.main.activity_scalebar.*
+
+/**
+ * Activity showing a scalebar used on mapview.
+ */
+class ScalebarActivity : AppCompatActivity() {
+    private var scaleBar: ScaleBar? = null
```
tobrun: when using `import kotlinx.android.synthetic.main.activity_scalebar.*` you can reference your scaleBar directly in your code, no need to add a field for it.

tobrun: this also removes the requirement of needed to add `scaleBar!!` all over the code :)

plugin-scalebar/src/androidTest/java/com/mapbox/pluginscalebar/ExampleInstrumentedTest.java
```diff
@@ -0,0 +1,26 @@
...
+@RunWith(AndroidJUnit4.class)
+public class ExampleInstrumentedTest {
+  @Test
+  public void useAppContext() {
+    // Context of the app under test.
+    Context appContext = InstrumentationRegistry.getTargetContext();
+
+    assertEquals("com.mapbox.pluginscalebar.test", appContext.getPackageName());
+  }
+}
```
tobrun: nit autogenerated template file, can be removed

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBar.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @Override
+  public void onScaleBegin(@NonNull StandardScaleGestureDetector detector) {
+
+  }
+
+  /**
+   * Set colors for the scale bar.
+   *
+   * @param textColor      The color for the texts above scale bar.
```
tobrun: nit above `the` scale bar

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBar.java
```diff
@@ -0,0 +1,92 @@
...
+import com.mapbox.android.gestures.StandardScaleGestureDetector;
+import com.mapbox.mapboxsdk.camera.CameraPosition;
+import com.mapbox.mapboxsdk.maps.MapView;
+import com.mapbox.mapboxsdk.maps.MapboxMap;
+import com.mapbox.mapboxsdk.maps.Projection;
+
+/**
+ * Plugin class that show scale bar on mapView and changes scale corresponding to mapView's scale.
+ */
+public class ScaleBar implements MapboxMap.OnScaleListener {
```
tobrun: Implementing `MapboxMap.OnScaleListener` listener on top of this public class also adds `onScaleBegin`, `onScaleEnd` and `onScale` to the public API. It would be nice if we could hide those by using an inner class instead.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.Locale;
+
+/**
+ * The scale widget is a visually representation of the scale plugin.
+ */
+class ScaleBarWidget extends View {
+  private int textColor;
+  private int primaryColor;
+  private int secondaryColor;
+  private int width;
```
tobrun: the width field is actually the mapView width as seen in https://github.com/mapbox/mapbox-plugins-android/pull/955/files#diff-b120ab4a1074a16f4fbb0d58dd43150dR106. Renaming it to `mapViewWidth` (and the setter) makes that instantly clear to the reader of the code.

plugin-scalebar/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="app_name">pluginScalebar</string>
```
tobrun: should be removed as this can trigger lint warnings in consuming apps.

plugin-scalebar/src/test/java/com/mapbox/pluginscalebar/ExampleUnitTest.java
```diff
@@ -0,0 +1,17 @@
...
+ * Example local unit test, which will execute on the development machine (host).
+ *
+ * @see <a href="http://d.android.com/tools/testing">Testing documentation</a>
+ */
+public class ExampleUnitTest {
+  @Test
+  public void addition_isCorrect() {
+    assertEquals(4, 2 + 2);
+  }
+}
```
tobrun: nit autogenerated template file, can be removed

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_scalebar)
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync { mapboxMap ->
+            mapboxMap.setStyle(Style.MAPBOX_STREETS) { _ ->
```
langsmith: ` _ ->` can be removed

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.os.Bundle
+import android.support.v7.app.AppCompatActivity
+import com.mapbox.mapboxsdk.camera.CameraUpdateFactory
+import com.mapbox.mapboxsdk.maps.Style
+import com.mapbox.mapboxsdk.plugins.testapp.R
+import com.mapbox.pluginscalebar.ScaleBar
+import kotlinx.android.synthetic.main.activity_scalebar.*
+
+/**
+ * Activity showing a scalebar used on mapview.
```
langsmith:
```suggestion
 * Activity showing a scalebar used on a MapView.
```

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBar.java
```diff
@@ -0,0 +1,92 @@
...
+import android.view.View;
+
+import com.mapbox.android.gestures.StandardScaleGestureDetector;
+import com.mapbox.mapboxsdk.camera.CameraPosition;
+import com.mapbox.mapboxsdk.maps.MapView;
+import com.mapbox.mapboxsdk.maps.MapboxMap;
+import com.mapbox.mapboxsdk.maps.Projection;
+
+/**
+ * Plugin class that show scale bar on mapView and changes scale corresponding to mapView's scale.
```
langsmith:
```suggestion
 * Plugin class that shows a scale bar on MapView and changes the scale corresponding to the MapView's scale.
```

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,161 @@
...
+import android.graphics.Paint;
+import android.support.annotation.ColorInt;
+import android.util.Pair;
+import android.view.View;
+
+import java.util.ArrayList;
+import java.util.Locale;
+
+/**
+ * The scale widget is a visually representation of the scale plugin.
```
langsmith:
```suggestion
 * The scale widget is a visual representation of the scale bar plugin.
```

Chaoba:
@tobrun When using double click to zoom in, `ScaleListener` can't get notified, how can we handle this scenario?

LukasPaczos:
@Chaoba instead of using `MapboxMap.OnScaleListener` I think it'd better to use `OnCameraMoveListener` and invalidate the view whenever zoom or latitude changes. Additionally, it'd be great to cache camera values and invalidate the view in loops, every `x` milliseconds (if the camera position value has changed) instead of on each update which can be extremely frequent.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,214 @@
...
+   */
+  private static class RefreshHandler extends Handler {
+    WeakReference<ScaleBarWidget> scaleBarWidgetWeakReference;
+
+    RefreshHandler(ScaleBarWidget scaleBarWidget) {
+      scaleBarWidgetWeakReference = new WeakReference<>(scaleBarWidget);
+    }
+
+    public void handleMessage(Message msg) {
+      if (msg.what == MSG_WHAT && scaleBarWidgetWeakReference.get() != null) {
```
LukasPaczos: Could you move `scaleBarWidgetWeakReference.get()` up, out of the if-check, and store a reference in a local variable? Otherwise, you're still in risk of running into a race condition where the reference is valid during the if-check, but not during the access.

Chaoba: Good point.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBar.java
```diff
@@ -0,0 +1,91 @@
...
+   * @param secondaryColor The color for even number index bars.
+   */
+  public void setColors(@ColorRes int textColor, @ColorRes int primaryColor, @ColorRes int secondaryColor) {
+    scaleBarWidget.setColors(textColor, primaryColor, secondaryColor);
+  }
+
+  class CameraMoveListener implements MapboxMap.OnCameraMoveListener {
+    @Override
+    public void onCameraMove() {
+      CameraPosition cameraPosition = mapboxMap.getCameraPosition();
```
LukasPaczos: The camera position should be cached, and `ScaleBarWidget#setDistancePerPixel` should only be called if `previousCamera.zoom != currentCamera.zoom` or `previousCamera.latitude != currentCamera.latitude`. Additionally, I think it'd be great to include thresholds, especially for latitude as it's not as sensitive as zoom level when it comes to the scale. For example, update only if `abs(previousCamera.latitude - currentCamera.latitude) > x`. This will help to limit the scale view updates even more.

Chaoba: How can we choose this thresholds?

LukasPaczos: @Chaoba you can verify how meters per pixel change based on latitude in our projection and select a threshold based on that. You will definitely need to update more frequently the closer you get to the poles.

Chaoba: @LukasPaczos I have do a test and get the following conclusions:
1. `MetersPerPixelAtLatitude` changes more and more often when getting to the poles, and the logic will be complicated to choose these thresholds.
2. Currently  every time we zoom in/out or move camera, it will draw 1~2 times.

So IMHO, we shouldn't pay too much efforts to import so complicated logic for this tiny issue and take the risk of having potential bugs.

LukasPaczos: Alright, thanks @Chaoba.
> Currently every time we zoom in/out or move camera, it will draw 1~2 times

I think this is acceptable, we can revisit with next iterations if needed.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,215 @@
...
+    super(context);
+    this.context = context;
+    textPaint.setAntiAlias(true);
+    textPaint.setTextSize(20);
+    textPaint.setTextAlign(Paint.Align.CENTER);
+    barPaint.setAntiAlias(true);
+    textColor = context.getResources().getColor(android.R.color.black);
+    primaryColor = context.getResources().getColor(android.R.color.black);
+    secondaryColor = context.getResources().getColor(android.R.color.white);
+    isMetricSystem = LocaleUnitResolver.isMetricSystem();
```
LukasPaczos: Could you expose a setter for `isMetricSystem`, besides loading a default value from `Locale`?

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,215 @@
...
+  private ArrayList<Pair<Integer, Integer>> scaleTable;
+  private String unit;
+  private RefreshHandler refreshHandler;
+  private Context context;
+
+  public ScaleBarWidget(Context context) {
+    super(context);
+    this.context = context;
+    textPaint.setAntiAlias(true);
+    textPaint.setTextSize(20);
```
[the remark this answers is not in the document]
LukasPaczos: This'd be great to expose as well.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,215 @@
...
+    float unitBarWidth = maxBarWidth / 2f;
+    if (unitDistance == 0) {
+      unitDistance = 1;
+    } else {
+      unitBarWidth = (float) (unitDistance / distancePerPixel);
+    }
+    //Drawing the surrounding borders
+    barPaint.setStyle(Paint.Style.FILL_AND_STROKE);
+    barPaint.setColor(secondaryColor);
+    int barHeight = 10;
```
LukasPaczos: This should also come from dimensions. Value can be cached after object construction and reused here.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -0,0 +1,215 @@
...
+    int borderWidth = 2;
```
LukasPaczos: Same as above.

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        mapView.onDestroy()
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        mapView?.onSaveInstanceState(outState)
```
LukasPaczos: NIT: unnecessary `?`.

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,65 @@
...
+ * Activity showing a scalebar used on a MapView.
+ */
+class ScalebarActivity : AppCompatActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_scalebar)
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync { mapboxMap ->
+            mapboxMap.setStyle(Style.MAPBOX_STREETS) {
+                mapboxMap.moveCamera(CameraUpdateFactory.zoomTo(2.0))
```
LukasPaczos: NIT: Do we need this invocation? Initially looked to me like the state saving was broken.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBar.java
```diff
@@ -0,0 +1,91 @@
...
+
+  public ScaleBar(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap) {
+    this.mapboxMap = mapboxMap;
+    this.projection = mapboxMap.getProjection();
+    this.scaleBarWidget = new ScaleBarWidget(mapView.getContext());
+    mapView.addView(scaleBarWidget);
+    cameraMoveListener = new CameraMoveListener();
+    mapboxMap.addOnCameraMoveListener(cameraMoveListener);
+    scaleBarWidget.setMapViewWidth(mapView.getWidth());
+  }
```
LukasPaczos: The logic from `onCameraMove` will need to be invoked here to query the meters per pixel and invalidate the view. Otherwise, on an initial start, the bar is not going to show up, because the camera has not moved yet.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarManager.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.mapbox.mapboxsdk.camera.CameraPosition;
+import com.mapbox.mapboxsdk.maps.MapView;
+import com.mapbox.mapboxsdk.maps.MapboxMap;
+import com.mapbox.mapboxsdk.maps.Projection;
+
+/**
+ * Plugin class that shows a scale bar on MapView and changes the scale corresponding to the MapView's scale.
+ */
+public class ScaleBarManager {
```
LukasPaczos: Why not a `ScaleBarPlugin`?

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarOption.java
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  /**
+   * Set the margin between scale bar and the top of mapView.
+   *
+   * @param marginTop the margin between scale bar and the top of mapView, in dp.
+   * @return this.
+   */
+  public ScaleBarOption setMarginTop(@DimenRes int marginTop) {
+    this.marginTop = convertDpToPixel(context.getResources().getDimension(marginTop));
```
LukasPaczos: No need to call `convertDpToPixel`. `getDimension` call already returns the value in pixels.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarOption.java
```diff
@@ -0,0 +1,307 @@
...
+
+  /**
+   * Set the secondary color of the scale bar, will be used to draw even index blocks,
+   * default value is android.R.color.white.
+   *
+   * @param secondaryColor the secondaryColor color of the scale bar.
+   * @return this.
+   */
+  public ScaleBarOption setSecondaryColor(@ColorRes int secondaryColor) {
+    this.secondaryColor = context.getResources().getColor(secondaryColor);
```
LukasPaczos: This method is deprecated in API 23. It's better to use `ContextCompat#getColor`.

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/scalebar/ScalebarActivity.kt
```diff
@@ -0,0 +1,82 @@
...
+        mapView.onCreate(savedInstanceState)
+        mapView.getMapAsync { mapboxMap ->
+            mapboxMap.setStyle(Style.MAPBOX_STREETS) {
+                addScalebar(mapboxMap)
+            }
+        }
+    }
+
+    private fun addScalebar(mapboxMap: MapboxMap) {
+        val scaleBarManager = ScaleBarPlugin(mapView, mapboxMap)
```
langsmith:
```suggestion
        val scaleBarPlugin = ScaleBarPlugin(mapView, mapboxMap)
```

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarOption.java
```diff
@@ -0,0 +1,296 @@
...
+import android.support.annotation.DimenRes;
+import android.support.annotation.VisibleForTesting;
+import android.support.v4.content.ContextCompat;
+
+import java.util.Locale;
+
+/**
+ * Builder class from which a scale bar is created.
+ */
+public class ScaleBarOption {
```
langsmith:
```suggestion
public class ScaleBarOptions {
```

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarPlugin.java
```diff
@@ -0,0 +1,91 @@
...
+    this.projection = mapboxMap.getProjection();
+  }
+
+  /**
+   * Create a scale bar widget on mapView.
+   *
+   * @param option The scale bar widget options that used to build scale bar widget.
+   * @return The created ScaleBarWidget instance.
+   */
+  public ScaleBarWidget create(ScaleBarOptions option) {
```
LukasPaczos: Missing `@NonNull` annotation.

Chaoba:
@LukasPaczos Thanks for your help. 🚀

virtuoso80:
Tried it works and looks great. It would be nicer though if you could define a fixed horizontal size for the Scale Bar so that it doesn't constantly resize itself. Thanks

LukasPaczos:
Hey @virtuoso80, thanks for the feedback! Could you cut a separate ticket for your request? Thank you!

Issue #817, opened 2019-01-13
NullPointerException in PlacePickerActivity


abustany:
I'm getting this crash in PlacePickerActivity while doing some panning and zooming (I couldn't reproduce the crash afterwards, but I was panning at pretty random locations):

```
java.lang.NullPointerException: Attempt to invoke virtual method 'double com.mapbox.mapboxsdk.geometry.LatLng.getLongitude()' on a null object reference
    at com.mapbox.mapboxsdk.plugins.places.picker.ui.PlacePickerActivity.makeReverseGeocodingSearch(PlacePickerActivity.java:178)
    at com.mapbox.mapboxsdk.plugins.places.picker.ui.PlacePickerActivity.onCameraIdle(PlacePickerActivity.java:159)
    at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher.executeOnCameraIdle(CameraChangeDispatcher.java:140)
    at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher.access$300(CameraChangeDispatcher.java:21)
    at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher$CameraChangeHandler.handleMessage(CameraChangeDispatcher.java:169)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:164)
    at android.app.ActivityThread.main(ActivityThread.java:6494)
    at java.lang.reflect.Method.invoke(Method.java)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
```

Full info on Sentry: https://sentry.io/share/issue/6698b0efe60643ddb91a5f3d56f4d62d/

langsmith:
Thanks for using Mapbox and the ticket , @abustany 

https://sentry.io/share/issue/6698b0efe60643ddb91a5f3d56f4d62d/ is dead for me

abustany:
Hmm it looks like Sentry events expire after a while... I don't see it in sentry anymore either. As far as I can remember there was not much more info on the Sentry page, the main thing is the stack trace, which we luckily still have :) I still have the email from Sentry, the only relevant piece of info I can see there is that the phone was running Android 8.1.0.

langsmith:
Ok, thanks @abustany . Just landed #964 which should solve this crash.

PR #964, opened 2019-05-22
Added check for null latLng in PlacePickerActivity makeReverseGeocodingSearch()

langsmith:
Resolves https://github.com/mapbox/mapbox-plugins-android/issues/817 by checking whether `latLng` from  `LatLng latLng = mapboxMap.getCameraPosition().target;` is null. Null `latLng` seemed to be the issue reported in #817 .


langsmith:
Not sure whether anything else is needed in this pr. Throw some sort of Exception message? Or no need?

PR #965, opened 2019-05-22
Adding includeReverseGeocoding call method to PlacePickerOptions

### LINKED ISSUE #857: Place picker without geocoding option
We want a simple location picker, just to return back lat/lng, without Geocoding option. Is it possible with current API-s?

langsmith:
This pr resolves #857 by adding a `PlacePickerOptions.Builder()` method to `includeReverseGeocode(boolean)`. If `false` is passed through, the bottom sheet is hidden and _no geocoding call is made_ (saving developer $). If `true`, bottom sheet is shown with reverse geocoding information displayed. 

cc @kbauhaus as fyi about search-related functionality.

ezgif com-resize (3)

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/places/PickerLauncherActivity.kt
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/places/PickerLauncherActivity.kt
@@ -4,6 +4,7 @@ import android.app.Activity
 import android.content.Intent
 import android.os.Bundle
 import android.support.v7.app.AppCompatActivity
+import android.widget.Switch
 import android.widget.Toast
 
 import com.mapbox.mapboxsdk.Mapbox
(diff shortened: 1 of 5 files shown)


langsmith:
cc @androideveloper , who asked about this in #857

langsmith:
Also meant to note that if `false` is passed through, `PlacePicker.getLastCameraPosition(data)` should be used to get the `CameraPosition` that represents the picked location. This is done in the test app's `PickerLauncherActivity`: 

https://github.com/mapbox/mapbox-plugins-android/blob/acf7a5da2b6e51183642bc4778ce373f01901121/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/places/PickerLauncherActivity.kt#L54-L57

androideveloper:
Great, Thanks! Btw, is there a way to enable "My location" button in Place Picker page that zooms to user current location?

androideveloper:
cc @karapetyanmery, take a look

langsmith:
@androideveloper ,

> is there a way to enable "My location" button in Place Picker page that zooms to user current location

There currently isn't a way to enable this, but it wouldn't be too hard to add this. I've created https://github.com/mapbox/mapbox-plugins-android/issues/973 to track your request.

PR #968, opened 2019-05-29
Combining the place picker and the geocode UI address autocomplete

### LINKED ISSUE #818: Widget combining the place picker and the address autocompleter
The place picker from Google shows a search field on top of the map (see https://developers.google.com/places/android-sdk/placepicker ), which allows to search for a place by its name first, and potentially refine the position afterwards by panning and zooming. In Mapbox, those are two separate widgets (or two activities you can start and wait for). I'm sure combining the two widgets from the Mapbox SDK together in a new activity is no rocket science, but it'd be convenient to have it as a built-in so I could just invoke the activity and retrieve the CarmenFeature as a result... Maybe having a search field could just be an option enabled via a field in `PlacePickerOptions` ?

langsmith:
Resolves #818 by combining the Place Picker and geocoding autocomplete UI functionalities together. Several folks in #818  have expressed interest in having this. 

@mapbox/maps-android , what do you think about the code and UI so far?

ezgif com-resize (6)


LukasPaczos:
Looks cool 👍 I'd add the searchbox to the toolbar and collapse the search results list when a record is clicked.

langsmith:
Made a little progress on this today.


`.includeSearch(false)` leads to the instructional `Choose a location` toolbar text:
ezgif com-resize

`.includeSearch(true)` places the autocomplete search UI widget in the toolbar area:

ezgif com-resize (1)


The next TODO is hiding the search UI dropdown when a location is selected.

percula:
This is great! Any chance of this being merged by the 29th, when Google officially deprecates their PlacePicker?

langsmith:
Ok, things are finally behaving. This is ready for some more review. 

The results drop-down list appears (i.e. the `SearchView` height is lengthened) when:

- typing occurs in the search UI `EditText`. Can be either the first text entered in the `EditText` or typing to adjust existing text.
- focus switches on to the search UI `EditText`. This happens when someone taps on the `EditText` for the first time and the cursor starts blinking in the `EditText` field.


The results drop-down list disappears (i.e. the `SearchView` height is shortened) when:

- it's showing and the map is tapped on.
- a location in the results list is tapped on.
- there's text in the search UI `EditText` and then the `X` clear button is tapped.

ezgif com-resize (9)

ezgif com-resize (6)

ezgif com-resize (7)

LukasPaczos:
A couple of notes:

1. The view is rescaled improperly when a location is selected from the dropdown list (it cuts the toolbar). The search view should also lose focus and collapse when an item is clicked and the keyboard should be hidden.
ezgif com-video-to-gif (7)

2. When the search view is in focus and extended it should lose focus and be hidden on back button click instead of closing the activity.
ezgif com-video-to-gif (6)

3. The `MapView` is covered by the toolbar which, for example, hides the compass. The map should either have top padding equal to the base height of the toolbar (to push the compass down) or be rescaled entirely.
device-2019-08-08-113026

4. The status bar is black while it should use the primary dark color.

5. The back arrow should be white as in other activities. 

6. Font color of the search view could be changed to something more visible (also white?).

7. When the search view is in focus for the first time, with no cache, it just extends a blank page. It should be hidden until the first results are available.
ezgif com-video-to-gif (8)

langsmith:
> The status bar is black while it should use the primary dark color.

ezgif com-resize (8)

> The back arrow should be white as in other activities. 
> 
> 
> Font color of the search view could be changed to something more visible (also white?).

[White is the toolbar's default background color](https://github.com/mapbox/mapbox-plugins-android/blob/master/plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/autocomplete/model/PlaceOptions.java#L202). This is why I've kept the back arrow, clear `X` button, and the search (hint) font colors as navy blue. I'm assuming that most people will keep it as default. I've added a check in `PlaceAutocompleteFragment` to check if the toolbar color **isn't** white. If it's not white, the back arrow, clear `X` button, and the search (hint) font are switched to white.

In the GIF below, you'll see one example using a white toolbar and another one using a red toolbar. Toolbar color detection and adjustments look to be working fine in regular `PlaceAutocompleteFragment` usage and in combined placepicker + search mode.

ezgif com-resize (10)


> When the search view is in focus and extended it should lose focus and be hidden on back button click instead of closing the activity.

Back button now hides the keyboard and removes focus from the `SearchView`. If in search + place combo mode, then the `SearchView` height is also collapsed:

ezgif com-resize (11)

> When the search view is in focus for the first time, with no cache, it just extends a blank page. It should be hidden until the first results are available

GIF below shows a fresh install. Clicking on the `SearchView`doesn't lead to expansion because `searchHistoryEntities.size()` is 0. After a place is selected and you return to place picker, clicking on `EditText` leads to a drop down because there's an entry in search history. 

ezgif com-resize (12)

> The `MapView` is covered by the toolbar which, for example, hides the compass. The map should either have top padding equal to the base height of the toolbar (to push the compass down) or be rescaled entirely.

Top padding of the map is now adjusted based on whether the `SearchView` drop down is expanded or not.

ezgif com-resize (17)

ezgif com-resize (18)


> The view is rescaled improperly when a location is selected from the dropdown list (it cuts the toolbar). The search view should also lose focus and collapse when an item is clicked and the keyboard should be hidden.

Focus is cleared and keyboard disappears. Toolbar height is now set to action bar size, rather than a hardcoded value. This should be good for dealing with various devices, displays, etc. 

There's still something weird happening with the device location button when a location is selected. Will look into it.

ezgif com-resize (15)

ezgif com-resize (16)

langsmith:
https://github.com/mapbox/mapbox-plugins-android/pull/968/commits/0c64cf98a75eb3cf3f0d4f3470ac56e1e9caab0d shows the keyboard and puts focus on the search EditText when `PlaceAutocompleteFragment` is viewed. This resolves https://github.com/mapbox/mapbox-plugins-android/issues/1011 ( cc @jollygreenegiant, who requested this feature).

langsmith:
Ok, I fixed FAB behavior. This is ready for another round of review when you've got a chance and are in the mood.

langsmith:
ezgif com-resize
ezgif com-resize (1)
ezgif com-resize (2)

LukasPaczos:
ezgif com-video-to-gif
When the bottom sheet is hidden, the margin on the FAB is 0, while it should be 16dp or so. Also, when the bottom sheet is extended, the attribution view (`(i)` icon) does not match the padding of the logo.

ezgif com-video-to-gif (1)
Not sure what material guidelines are, but this slight movement of both fabs up and down does not look right. More of a nit pick this one, we can ship regardless.

ezgif com-video-to-gif (2)
There seems to be a race condition when extending the keyboard on view inflation that doesn't adjust the bottom sheets padding/margin correctly. Probably the keyboard measurement is performed too soon? Ultimately, do we even need the sheet to be adjusted for the keyboard?

ezgif com-video-to-gif (3)
If there's no search history and the keyboard is closed, the search view is still extended with an empty background. The search box generally doesn't lose focus on back press, while it probably should.

round_edges
Because the toolbar in search mode has round edges, it creates an unnatural gap between the status bar and toolbar.

PR #977, opened 2019-06-05
Fix get metersPerPixel twice error.

### LINKED ISSUE #972: ScaleBar incorrect values
Scale bar is out of sync with the map. 

Adding scalebar like so:

```
val scaleBar = ScaleBarPlugin(mapView, mapboxMap)
scaleBar.create(ScaleBarOptions(safeContext))
scaleBar.isEnabled = true
```

Result:

WhatsApp Image 2019-06-02 at 16 40 11

For reference see iOS version:
WhatsApp Image 2019-06-02 at 16 43 47

Chaoba:
Resolves #972 . 
Root cause for this issue is that we call `projection.getMetersPerPixelAtLatitude` twice and get the wrong value.


paulsUsername:
Looks perfect here now! Thanks

WhatsApp Image 2019-06-05 at 14 10 46

paulsUsername:
Looks perfect here now! Thanks

WhatsApp Image 2019-06-05 at 14 10 46

PR #984, opened 2019-06-10
Change the update strategy of marker view

### LINKED ISSUE #981: MakerView update delay
There are two phenomenons related to markerview update delay and could be reproduced by test app in this repo.
1. When open a new activity all markers will be firstly shown on the left top corner and then updated to their correct location.
2. When zooming out/in or sliding mapview, markers will be updated with a obvious delay.

Possible solution:
1. Call `markerView.update();` when add a new maker.
2. Instead of `OnDidFinishRenderingFrameListener`, use `OnCameraDidChangeListener` to update maker's position.

@tobrun What do you think?

Chaoba:
Resolves #981 
This PR fix update delay of marker view, the following pictures are the profile of GPU usage

Change before:
before
After change:
after

It seems the light green part which represent `Measure / Layout` increase, @tobrun do you think it is acceptable?

### DIFF
--- a/plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
+++ b/plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
@@ -12,7 +12,7 @@ import java.util.List;
 /**
  * Class responsible for synchronising views at a LatLng on top of a Map.
  */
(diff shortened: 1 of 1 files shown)


tobrun:
In those screenshots you aren't dropping any frames so this fits in the acceptable region

tobrun:
thinking about it, if the users want a slower synchronization they should be able to hook into your max fps feature to lower the amount of updates.

PR #992, opened 2019-06-18
Downgrade log level of MapLocale no match scenarios

snkashis:
Since not finding matches for MapLocale is a very common and normal occurrence, I don't believe this should be logged thru Timber as an error. Moved it to debug instead. This currently litters logs with huge amounts of duplicative information.


langsmith:
Looks good to me. @tobrun @LukasPaczos , you good with this too?

langsmith:
Closing in favor of https://github.com/mapbox/mapbox-plugins-android/pull/993

Thank you @snkashis

PR #997, opened 2019-06-21
Make LAYER_ID of BuildingPlugin public

snkashis:
This is helpful so other tools can reference it as an argument in `layerBelow` or `layerAbove`


LukasPaczos:
Thanks @snkashis! Could you add a javadoc entry as well?

snkashis:
@LukasPaczos ok done

PR #1000, opened 2019-06-28
Add automated resource loading of icon-images

### LINKED ISSUE #742: icon-image loader
The style-spec icon-image is currently exposed as 

```java
  public void setIconImage(String value) {
    jsonObject.addProperty("icon-image", value);
  }
```

We need an easier way to support Android resources, for example by allow setting a drawable, bitmap or resource id. This setup should also take in account to support re-adding the images if a style changes occurs #737.

tobrun:
This PR is a proposal to automate loading of icon-image property of symbols. If core misses an style image, it will emit the image missing event. The symbol manager will pick this up and will attempt to load a drawable from resources with this id. This way users [don't need to call addImage](https://github.com/mapbox/mapbox-plugins-android/compare/tvn-load-icons?expand=1#diff-34b94cbd412d63114950931bdb2c79bcL75) themselves, they can just pass in the id of the drawable resource as a String (see [here](https://github.com/mapbox/mapbox-plugins-android/compare/tvn-load-icons?expand=1#diff-34b94cbd412d63114950931bdb2c79bcR94)).

Closes #742 

@mapbox/maps-android

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/PressForSymbolActivity.java
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/PressForSymbolActivity.java
@@ -8,6 +8,7 @@ import android.os.Bundle;
 import android.support.annotation.DrawableRes;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
+import android.support.annotation.VisibleForTesting;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.app.AppCompatDelegate;
 import com.mapbox.mapboxsdk.camera.CameraPosition;
@@ -15,6 +16,7 @@ import com.mapbox.mapboxsdk.geometry.LatLng;
 import com.mapbox.mapboxsdk.maps.MapView;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
 import com.mapbox.mapboxsdk.maps.Style;
+import com.mapbox.mapboxsdk.plugins.annotation.AnnotationManager;
 import com.mapbox.mapboxsdk.plugins.annotation.SymbolManager;
 import com.mapbox.mapboxsdk.plugins.annotation.SymbolOptions;
 import com.mapbox.mapboxsdk.plugins.testapp.R;
@@ -145,6 +147,21 @@ public class PressForSymbolActivity extends AppCompatActivity {
       drawable.draw(canvas);
       return bitmap;
     }
+  }
+
+  @VisibleForTesting
+  public MapView getMapView(){
+    return mapView;
+  }
+
+  @VisibleForTesting
+  public SymbolManager getSymbolManager(){
+    return symbolManager;
+  }
+
+  @VisibleForTesting
+  public MapboxMap getMapboxMap(){
+    return mapboxMap;
   }
 }
 
--- a/plugin-annotation/scripts/annotation_manager.java.ejs
+++ b/plugin-annotation/scripts/annotation_manager.java.ejs
@@ -7,23 +7,30 @@
 
 package com.mapbox.mapboxsdk.plugins.annotation;
 
+import android.content.Context;
+import android.content.res.Resources;
+import android.graphics.Bitmap;
+import android.graphics.drawable.Drawable;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.UiThread;
 import android.support.annotation.VisibleForTesting;
 
+import android.support.v4.content.res.ResourcesCompat;
 import com.mapbox.geojson.Feature;
 import com.mapbox.geojson.FeatureCollection;
+import com.mapbox.mapboxsdk.log.Logger;
 import com.mapbox.mapboxsdk.maps.MapView;
 import com.mapbox.mapboxsdk.maps.MapboxMap;
+import com.mapbox.mapboxsdk.maps.OnMapReadyCallback;
 import com.mapbox.mapboxsdk.maps.Style;
 import com.mapbox.mapboxsdk.style.expressions.Expression;
-import com.mapbox.mapboxsdk.style.layers.<%- camelize(type) %>Layer;
-import com.mapbox.mapboxsdk.style.layers.PropertyValue;
-import com.mapbox.mapboxsdk.style.layers.PropertyFactory;
+import com.mapbox.mapboxsdk.style.layers.*;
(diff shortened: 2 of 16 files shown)


chloekraw:
@tobrun, this is awesome! @julianrex @fabian-guerra, is this also possible on the iOS side?

julianrex:
Yep, totally.

fabian-guerra:
@chloekraw my answer would be yes and no 😅

Unlike Android listeners approach. iOS uses a delegate architecture where the map can "listen" for events and pass this to its single delegate. This means if I implement the delegate in the annotations controller then the developer won't be able to use the map's delegate in his own classes. This is what I mean with "no": I won't be using the same approach. 

"Yes" means we can support it with a different approach.

plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -80,8 +87,62 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
...
+      this.mapViewWeakReference = new WeakReference<>(mapView);
+    }
+
+    @Override
+    public void onStyleImageMissing(@NonNull final String id) {
+      if (id.isEmpty()) {
+        return;
+      }
+
+      Logger.v(TAG, String.format("Style image missing %s", id));
```
LukasPaczos: We should decrease the level and change the message or skip this log, otherwise, it's going to pollute the logcat on each load.

plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -80,8 +87,62 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
...
+            Drawable drawable = ResourcesCompat.getDrawable(resources, drawableId, context.getTheme());
+            final Bitmap bitmap = BitmapUtils.getBitmapFromDrawable(drawable);
+            if (bitmap != null) {
+              mapView.getMapAsync(new OnMapReadyCallback() {
+                @Override
+                public void onMapReady(@NonNull MapboxMap mapboxMap) {
+                  mapboxMap.getStyle(new Style.OnStyleLoaded() {
+                    @Override
+                    public void onStyleLoaded(@NonNull Style style) {
+                      Logger.v(TAG, String.format("Image lookup success, adding image missing %s", id));
```
LukasPaczos: Same as above.

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/SymbolManager.java
```diff
@@ -96,6 +103,56 @@ public SymbolManager(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap, @No
   @VisibleForTesting
   SymbolManager(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap, @NonNull Style style, @NonNull CoreElementProvider<SymbolLayer> coreElementProvider, @Nullable String belowLayerId, @Nullable GeoJsonOptions geoJsonOptions, DraggableAnnotationController<Symbol, OnSymbolDragListener> draggableAnnotationController) {
     super(mapView, mapboxMap, style, coreElementProvider, draggableAnnotationController, belowLayerId, geoJsonOptions);
+    mapView.addOnStyleImageMissingListener(new ImageLoader(mapView));
+  }
+
```
langsmith: It's pretty clear from the code, but perhaps add 2-3 javadoc sentences that explain what this class is about

app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/PressForSymbolActivity.java
```diff
@@ -146,5 +148,20 @@ static Bitmap getBitmapFromDrawable(Drawable drawable) {
       return bitmap;
     }
   }
+
+  @VisibleForTesting
```
LukasPaczos: Can changes here be reverted? They seem unused.

plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -74,12 +81,15 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
    */
   @UiThread
   public <%- camelize(type) %>Manager(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap, @NonNull Style style, @Nullable String belowLayerId, @Nullable GeoJsonOptions geoJsonOptions) {
-    this(mapView, mapboxMap, style, new <%- camelize(type) %>ElementProvider(), belowLayerId, geoJsonOptions, new DraggableAnnotationController<<%- camelize(type) %>, On<%- camelize(type) %>DragListener>(mapView, mapboxMap));
+    this(mapView, mapboxMap, style, new <%- camelize(type) %>ElementProvider(), belowLayerId, geoJsonOptions, new DraggableAnnotationController<<%- camelize(type) %>, On<%- camelize(type) %>DragListener>(mapView, mapboxMap), new ImageLoader(mapView));
```
LukasPaczos: The `ImageLoader` argument addition should be if-checked for symbol manager.

plugin-annotation/scripts/annotation_manager.java.ejs
```diff
@@ -74,12 +81,15 @@ public class <%- camelize(type) %>Manager extends AnnotationManager<<%- camelize
...
   }
 
   @VisibleForTesting
-  <%- camelize(type) %>Manager(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap, @NonNull Style style, @NonNull CoreElementProvider<<%- camelize(type) %>Layer> coreElementProvider, @Nullable String belowLayerId, @Nullable GeoJsonOptions geoJsonOptions, DraggableAnnotationController<<%- camelize(type) %>, On<%- camelize(type) %>DragListener> draggableAnnotationController) {
+  <%- camelize(type) %>Manager(@NonNull MapView mapView, @NonNull MapboxMap mapboxMap, @NonNull Style style, @NonNull CoreElementProvider<<%- camelize(type) %>Layer> coreElementProvider, @Nullable String belowLayerId, @Nullable GeoJsonOptions geoJsonOptions, DraggableAnnotationController<<%- camelize(type) %>, On<%- camelize(type) %>DragListener> draggableAnnotationController, ImageLoader imageLoader) {
```
LukasPaczos: Same as above.

oantajames:
@tobrun any updates on this? 😄

PR #1010, opened 2019-07-11
Change the unit of scalebar according distance

### LINKED ISSUE #1009: Support showing distance in km vs meters
Capturing from a customer that they are seeing values as 1500m on Android scale widget implementation while they see 1.5 km on the iOS one. We should try to align this more or expose a configuration to achieve this effect.

cc @Chaoba @chloekraw

Chaoba:
Resolves #1009. 
Unit will change to km if current unit is meter and distance > 1000m, and for ft it will change to mile if distance > 5280ft.

### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
+++ b/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
@@ -6,9 +6,12 @@ import java.util.ArrayList;
 
 class ScaleBarConstants {
   static double FEET_PER_METER = 3.2808;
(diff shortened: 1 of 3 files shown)


plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
```diff
@@ -6,9 +6,12 @@
 
 class ScaleBarConstants {
   static double FEET_PER_METER = 3.2808;
+  static int KILO_METER = 1000;
```
tobrun: kilometer is one word

tobrun:
cc @chloekraw

PR #1028, opened 2019-08-06
Expand the max distance of scale bar.

Chaoba:
The max distance for scale bar is not large enough, so the distance text will overlap at zoom level 0. Expand the max distance to make scale bar display normally at zoom level 0.
device-2019-08-06-152824
device-2019-08-06-152807
CC: @chloekraw

### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
+++ b/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
@@ -41,6 +41,10 @@ class ScaleBarConstants {
       add(new Pair<>(500000, 2));
       add(new Pair<>(600000, 3));
       add(new Pair<>(800000, 2));
+      add(new Pair<>(1000000, 2));
+      add(new Pair<>(2000000, 2));
+      add(new Pair<>(3000000, 3));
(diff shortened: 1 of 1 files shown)


LukasPaczos:
Looks good :+1: I was just wondering if you could add even more steps to future proof the impl for constantly increasing screen resolutions. For example, I can reach over the 3000 miles step, up until 5000 miles on a Pixel 2 XL.

Chaoba:
@LukasPaczos That makes sense, will add more later.

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarConstants.java
```diff
@@ -41,6 +41,13 @@
       add(new Pair<>(500000, 2));
       add(new Pair<>(600000, 3));
       add(new Pair<>(800000, 2));
+      add(new Pair<>(1000000, 2));
+      add(new Pair<>(2000000, 2));
+      add(new Pair<>(3000000, 3));
+      add(new Pair<>(4000000, 2));
+      add(new Pair<>(5000000, 2));
+      add(new Pair<>(6000000, 3));
+      add(new Pair<>(8000000, 2));
```
LukasPaczos: The metric unit can go to `10000000` or even further.

PR #1048, opened 2019-09-04
Add option to config max width of scalebar.

### LINKED ISSUE #1046: Add fixed width option to scale bar
Tried the scale bar and it works and looks great. It would be nicer though if you could define a fixed horizontal size for the Scale Bar so that it doesn't constantly resize itself. The auto resize sometimes takes up half the screen.
Thanks

Chaoba:
Resolves #1046 
ratio is calculated according scalebar with / mapview width.
Currently we add support for HALF, THIRD and QUATER, may add more if needed in future.

### DIFF
--- a/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
+++ b/plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
@@ -19,6 +19,9 @@ import static com.mapbox.pluginscalebar.ScaleBarConstants.FEET_PER_MILE;
 import static com.mapbox.pluginscalebar.ScaleBarConstants.KILOMETER;
 import static com.mapbox.pluginscalebar.ScaleBarConstants.KILOMETER_UNIT;
 import static com.mapbox.pluginscalebar.ScaleBarConstants.MILE_UNIT;
+import static com.mapbox.pluginscalebar.ScaleBarOptions.ScaleBarRatio.HALF;
+import static com.mapbox.pluginscalebar.ScaleBarOptions.ScaleBarRatio.QUARTER;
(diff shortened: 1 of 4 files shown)


Chaoba:
I have the same thoughts at first, the current implementation is based on the following reasons:
1. The defined values is more clear and straightforward.
2. No need to add checks to insure the passed in param is valid. 
3. Prevent the corner scenarios like text overlapping.

tobrun:
> The defined values is more clear and straightforward.

Being able to define a maxWidth configuration will be more flexible as it will cater to all use-cases / designs. I feel, if a user wants a scalebar of max 30 pixels, they should be able to configure that.

Chaoba:
@tobrun Consider of adaption for difficult devices, I believe use ratio is better than pixels. So, how about we let user set the a ratio in 0.1 ~ 1?

tobrun:
> Consider of adaption for difficult devices, I believe use ratio is better than pixels. 

Android view system is build up around pixels. To support variety of device screen sizes, we do have the concept of dp values but that is only used in xml resources. In code you always use pixels. I'm personally not seeing with using pixels.

virtuoso80:
Hi, any update?

plugin-scalebar/src/main/java/com/mapbox/pluginscalebar/ScaleBarWidget.java
```diff
@@ -363,6 +364,22 @@ private String getDistanceText(int distance) {
     }
   }
 
+  /**
+   * Set the ratio of scale bar max width compared with MapView width.
+   * @param ratio the ratio scale bar will use.
+   */
+  public void setRatio(float ratio) {
```
tobrun: Maybe rename to `setMaxWidthRatio` , ratio itself would mean the screen pixel ratio

tobrun:
@Chaoba can you resolve conflicting files, let's merge as is and revisit the pixel configuration later.

Issue #1102, opened 2020-03-05
Release new versions of plugins with v9 prefix


langsmith:
This ticket is to track the update releases of **all** of our plugins. Various features/fixes have been added since their last release, most notably the adoption of [AndroidX](https://developer.android.com/jetpack/androidx). 

The plugin module names will now have the `v9` prefix instead of `v8` because [`9.0.0` was the last semver major release of the Mapbox Maps SDK for Android](https://github.com/mapbox/mapbox-gl-native-android/releases/tag/android-v9.0.0).

##### Features
- Switching all plugins to AndroidX #1100
- Added user location button to Places Plugin's PlacePickerActivity #994

##### Features
- Switching all plugins to AndroidX #1100

##### Features
- Switching all plugins to AndroidX #1100

##### Bugs
- Switch Timber log level to debug for setMapLanguage #1085

##### Features
- Switching all plugins to AndroidX #1100
- Expose underlying, generated Layer ID #1071

##### Features
- Switching all plugins to AndroidX #1100

##### Features
- Switching all plugins to AndroidX #1100
- Make LAYER_ID of BuildingPlugin public #997

##### Features
- Switching all plugins to AndroidX #1100

##### Bugs
- Fix scale bar incorrect scale #1088

##### Features
- Switching all plugins to AndroidX #1100

## Release checklist
- [x] Cut the release branch
- [x] Run through test/demo app and test all activities
- [x] Update `CHANGELOG.md`
- [x] Change version names in `gradle.properties`
- [x] Publish artifacts to Maven Central
- [x] Change version name back to `SNAPSHOT`
- [x] Merge the release branch back to `master`
- [x] Tag releases in GitHub

## Post-release @langsmith 
- [x] Update `README.md`
- [x] Update `constants.js` file in `android-docs` repo 
- [x] Generate updated javadocs and add to `android-docs` repo 
- [x] If needed, bump dependencies in https://github.com/mapbox/mapbox-android-demo

/cc: @mapbox/android @mapbox/enterprise-support @mapbox/maps-pms

langsmith:
Done here!

![giphy](https://user-images.githubusercontent.com/4394910/76099513-11f5de80-5f80-11ea-865e-0b72253eb6ea.gif)

PR #1103, opened 2020-03-05
 Release new versions of all plugins with v9 prefix

langsmith:
This pr addresses https://github.com/mapbox/mapbox-plugins-android/issues/1102 by doing a new release of all of our plugins.


langsmith:
We're lookin' good on Sonatype
Screen Shot 2020-03-05 at 8 41 10 AM
Screen Shot 2020-03-05 at 8 41 20 AM
Screen Shot 2020-03-05 at 8 41 29 AM

Issue #908, opened 2019-04-04
Place Picker crashes on Android X


riastrad:
There is a known issue with the AndroidX migration tool in Android Studio that fails to update the XML for `BottomSheetBehavior` in projects. (See [Stack Overflow discussion](https://stackoverflow.com/questions/51617186/bottomsheetbehavior-not-in-androidx-libraries).)

Unfortunately our Places Plugin was not spared by this issue and will crash if you use the Place Picker on AndroidX.

<details>
  <summary>Example crash:</summary>
  <pre>
2019-04-03 11:26:47.146 16098-16098/com.your.androidapp E/AndroidRuntime: FATAL EXCEPTION: main
Process: com.your.androidapp, PID: 16098
java.lang.IllegalArgumentException: Illegal state argument: 5
at com.google.android.material.bottomsheet.BottomSheetBehavior.startSettlingAnimation(BottomSheetBehavior.java:1059)
at com.google.android.material.bottomsheet.BottomSheetBehavior.startSettlingAnimationPendingLayout(BottomSheetBehavior.java:846)
at com.google.android.material.bottomsheet.BottomSheetBehavior.setState(BottomSheetBehavior.java:825)
at com.mapbox.mapboxsdk.plugins.places.picker.ui.CurrentPlaceSelectionBottomSheet.toggleBottomSheet(CurrentPlaceSelectionBottomSheet.java:80)
at com.mapbox.mapboxsdk.plugins.places.picker.ui.CurrentPlaceSelectionBottomSheet.setPlaceDetails(CurrentPlaceSelectionBottomSheet.java:55)
at com.mapbox.mapboxsdk.plugins.places.picker.ui.PlacePickerActivity.onCameraIdle(PlacePickerActivity.java:158)
at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher.executeOnCameraIdle(CameraChangeDispatcher.java:140)
at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher.access$300(CameraChangeDispatcher.java:21)
at com.mapbox.mapboxsdk.maps.CameraChangeDispatcher$CameraChangeHandler.handleMessage(CameraChangeDispatcher.java:169)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:163)
at android.app.ActivityThread.main(ActivityThread.java:6238)
at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #1107, opened 2020-03-19
[places] Setting reverse geocoding BottomSheetBehavior hideable to true

langsmith:
Resolves https://github.com/mapbox/mapbox-plugins-android/issues/908 by adjusting the Places Plugin's BottomSheetBehavior.

cc @orlikraf, who had a helpful comment at https://github.com/mapbox/mapbox-plugins-android/issues/908#issuecomment-596203513

### DIFF
--- a/plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/picker/ui/CurrentPlaceSelectionBottomSheet.java
+++ b/plugin-places/src/main/java/com/mapbox/mapboxsdk/plugins/places/picker/ui/CurrentPlaceSelectionBottomSheet.java
@@ -77,7 +77,8 @@ public class CurrentPlaceSelectionBottomSheet extends CoordinatorLayout {
 
   private void toggleBottomSheet() {
     bottomSheetBehavior.setPeekHeight(rootView.findViewById(R.id.bottom_sheet_header).getHeight());
-    bottomSheetBehavior.setHideable(isShowing());
-    bottomSheetBehavior.setState(isShowing() ? STATE_HIDDEN : STATE_COLLAPSED);
+    boolean isShowing = isShowing();
+    bottomSheetBehavior.setHideable(isShowing);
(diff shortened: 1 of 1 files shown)


langsmith:
The crash that was happening

```
2020-03-19 00:16:06.986 9832-9832/? E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.mapbox.mapboxandroiddemo.debug, PID: 9832
    java.lang.IllegalArgumentException: Illegal state argument: 5
        at com.google.android.material.bottomsheet.BottomSheetBehavior.startSettlingAnimation(BottomSheetBehavior.java:1099)
```

Discussions about `java.lang.IllegalArgumentException: Illegal state argument: 5`:

- https://stackoverflow.com/a/47871634/6358488
- https://stackoverflow.com/a/53263494/6358488

langsmith:
ezgif com-resize

orlikraf:
> Resolves #908 by adjusting the Places Plugin's BottomSheetBehavior `setHideable()` value. I don't believe there's any downside to hardcoding the fact that the sheet can be hideable.
> 
> cc @orlikraf, who had a helpful comment at [#908 (comment)](https://github.com/mapbox/mapbox-plugins-android/issues/908#issuecomment-596203513)

@langsmith 
If I recall correctly a hideable BottomSheet can be dismissed out of screen with a swipe down without a way to swipe it into view again. I'm 90% sure about this so please check it yourself.

I would suggest to call `isShowing` only once before setting hideable and use the result in both later calls.

langsmith:
>I would suggest to call isShowing only once before setting hideable and use the result in both later calls.

Thanks @orlikraf , that's what I'm now doing in this pr. Getting a weird CI 🐛  now.

langsmith:
CircleCI 💥 is from a Mapbox infrastructure 🐛 , rather than something with this pr's code changes.

PR #1116, opened 2020-04-21
Added .baseUrl() to Places Plugin's PlaceOptions builder

langsmith:
This pr adds `.baseUrl()` to the Places Plugin's `PlaceOptions` builder. 

Using https://www.mapbox.com/atlas is a good example of when a user might want to make geocoding calls to a different base URL than the default Mapbox one.


langsmith:
CI tests aren't running because of known Github 🦄issues right now

Issue #883, opened 2019-03-18
Dragged annotation triggers Map Clicked


sebastienrouif:
Scenario :
 - on master "1e032afb0652582e19166ef41ab8db03ca064f33"
 - open the app in the CircleActivity, Drag a circle for a short amount of time and distance
 - observe a toast for CircleClicked

Expected behaviour :
 - as the circle has been dragged I was not expecting the click listener to be triggered

A similar problem happens with the long click

I'll try to investigate how the DraggableAnnotationController could be changed to include that behavior.

androideveloper:
The same here

JonasVautherin:
Is there a workaround to that? Like a way to detect in the long click callback that a drag is active?

ajaykotianneebal:
Any workaround found for this issue?

kikothemaster:
Any workaround found for this issue?

Issue #863, opened 2019-03-05
Order of initialization of annotation managers matters!


androideveloper:
I have found a weird bug! When you initialize symbol manager and the circle manager, your symbol can't be dragged! If you change the order of initialization it works!

### Steps to reproduce

Demo https://www.dropbox.com/s/ympu7zg722vehty/MapboxSymbolDragDemo.zip?dl=0

Try to change order of initialization:
 symbolManager = SymbolManager(mapView, mapboxMap, style)
 circleManager = CircleManager(mapView, mapboxMap, style)

### Expected behavior

Symbol should be draggable in any case

### Actual behavior

Symbol is draggable only, if circle manager initialized first

### Configuration

**Android versions:** 9.0
**Device models:** Xiaomi mi 2 lite
**Mapbox SDK versions:** 7.2.0, plugin-annotation-v7:0.5.0

LukasPaczos:
This is an interesting case, thanks for the report @androideveloper.

Because we are using `mapView.setOnTouchListener()` to intercept touch events, **only the last created manager will have draggable annotations**.

There seem to be 2 ways of resolving this:
- make `DraggableAnnotationController` a singleton that dispatches drag events to all managers
- expose a way to intercept map touch events before they are passed to the internal gestures manager. This would also resolve https://github.com/mapbox/mapbox-plugins-android/issues/884 as we wouldn't have to rely on the `OnTouchListener`.

Will update this ticket with progress, if we find any of above viable.

androideveloper:
@LukasPaczos Thanks for the update! I will keep it as the last created manager until this is resolved.

adrianababakanian:
We should re-open this, since recent [reports on Stack Overflow](https://stackoverflow.com/questions/60266329/android-problem-with-2-annotation-type-in-mapbox/60859256#60859256) indicate that developers are still encountering the issue.

m0nac0:
We're also encountering this in the mapbox_gl flutter plugin (tracked at https://github.com/tobrun/flutter-mapbox-gl/issues/130). This is kind of significant, as it means that we can only have one draggable annotation type.

m0nac0:
@adrianababakanian @LukasPaczos Sorry for spamming, but do you know if this is being worked on? Otherwise we'd have to find some workaround for the mapbox_gl flutter plugin to support draggable annotations.

virtuoso80:
[...]

PR #1125, opened 2020-05-06
[annotation] Update DraggableAnnotationController for multiple draggable managers (#863, #883)

Ph0tonic:
This is a proposal for fixing #863 and #883.

They are kind of linked together and I foud the solution of the second one during my work so I added it in this PR.

I know that it's not perfect but at least is a simple solution which allow multiple manager to be draggable.

@LukasPaczos and @langsmith I would be happy to have your feedback. 👍


LukasPaczos:
Hey @Ph0tonic, thanks for your contribution again!

Unfortunately, I don't think this is an approach that we'd like to take. Having multiple detectors running is going to get inefficient.

To resolve the issue, I'd rather explore the option of having one, public instance of the draggable controller into which we can inject multiple annotation managers, instead of each annotation manager creating its own draggable controller.

Ph0tonic:
Ok, I see, I will try to turn draggable controller into a singleton such has to minimize the number of detectors.

Ph0tonic:
@LukasPaczos Here is a completely new implementation which turns DraggableController into a singleton. Would this be closer to an approach you would take ?

Ph0tonic:
I had a look at the `make generate-annotation-code` and the linked scripts and I think that some of the `.ejs` needs to be updated but as these are annotated with:
```java
// This file is generated.
```

I let those like they currently are. Please let me known if I can do anything else @langsmith and @LukasPaczos. Thanks

Ph0tonic:
The limitations resolved by this PR would really help me for one of my side project. If we arrive to a solution, do you think we might be able to do a release for those next weeks ?
@LukasPaczos and @langsmith Thank you very much for your time.

kikothemaster:
When can we expect this fix in production?

Ph0tonic:
Any updates about this PR ? @LukasPaczos and @langsmith

plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/DraggableAnnotationController.java
```diff
@@ -16,17 +16,32 @@
...
+final class DraggableAnnotationController <T extends Annotation, D extends OnAnnotationDragListener<T>> {
+
+  private static DraggableAnnotationController INSTANCE = null;
+
+  public static synchronized DraggableAnnotationController getInstance(MapView mapView, MapboxMap mapboxMap) {
+    if(INSTANCE == null){
+      INSTANCE = new DraggableAnnotationController(mapView, mapboxMap);
+    }
+    return INSTANCE;
+  }
```
LukasPaczos, line 31: An issue with this singleton approach might occur when the map is recreated, for example in a recycler view, or when transitioning between fragments/activites.

We might need to do something like this, to additionally ensure that we're using the correct instance:
```suggestion
  public static synchronized DraggableAnnotationController getInstance(MapView mapView, MapboxMap mapboxMap) {
    if(INSTANCE == null || this.mapView != mapview || this.mapboxMap != mapboxMap){
      INSTANCE = new DraggableAnnotationController(mapView, mapboxMap);
    }
    return INSTANCE;
  }
```

The strong references to `mapView` and `mapboxMap` should be dropped when the last `removeAnnotationManager` is called and the instance is nullified so we shouldn't leak. It'd be great to test this scenario though.

Ph0tonic: You're right I did not think about that 👍. I updated the code to reflect what you said. Regarding the testing of the last part of your remark, I'm not sure how I can do it.

LukasPaczos:
Amazing work here, thanks @Ph0tonic! Let's continue in https://github.com/mapbox/mapbox-plugins-android/pull/1147.

Ph0tonic:
> 
> 
> Amazing work here, thanks @Ph0tonic! Let's continue in #1147.

You are welcome

PR #1129, opened 2020-05-18
Implement cancel draggable during onAnnotationDrag

Ph0tonic:
While using this plugin, I needed to cancel a drag event when my point was further apart from another location. The way I solved this was by deleting the point and repainting it so that cancel the drag event. The problem is that it's not efficient at all.

Here is a proposal to allow a drag event to be canceled directly in the `dragEventListener`. I would be pleased to have your feedback @langsmith and @LukasPaczos.

Thanks in advance

### DIFF
--- a/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/CircleActivity.java
+++ b/app/src/main/java/com/mapbox/mapboxsdk/plugins/testapp/activity/annotation/CircleActivity.java
@@ -103,12 +103,13 @@ public class CircleActivity extends AppCompatActivity {
         }
 
         @Override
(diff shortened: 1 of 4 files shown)


LukasPaczos:
Thanks @Ph0tonic, the idea and the implementation look good to me. Would you be able to add a unit test?

@langsmith could you help with merging this when ready?

Ph0tonic:
You're welcome. Hum yes I will try to add some test but I not sure anymore of my implementation. In the present case when a drag is cancel then the annotation stay still and if we continue our movement then the map moves. I think that this should not happend and could be very strange to the user has he was moving an annotation and sudently he is moving the map.

What do you think about cancelling the drag but still consuming the event such as to prevent the map to take back the event until the user end is drag ?

PR #1141, opened 2020-07-18
[annotation] Consume event when drag end to prevent click event

### LINKED ISSUE #883: Dragged annotation triggers Map Clicked

Ph0tonic:
I wrote a seperate PR for this issue such as to simplify the review of #1125.
- Fixes #883

Any review and comment would be much appreciated 👍 @LukasPaczos and @langsmith.


LukasPaczos:
Thanks @Ph0tonic!

Issue #189, opened 2017-12-01
Cancel download should remove region definition


tobrun:
Atm cancelling a download doesn't result in removing the created region for it. Since the download was cancelled, keeping the definition around isn't useful. You can reproduce this by downloading a region with the form, cancelling it and opening the region list activity.

djavid:
As of now canceling download do not delete region.

djavid:
@tobrun Can you please see my last comment at #12382

tobrun:
@djavid please do not reply to closed issues, those will get unnoticed. if you find a bug or have a feature request please open a new issue instead. If you need support, please contact us at https://www.mapbox.com/contact/ or through stackoverflow. That said, please add the appropriate context and information needed to make your issue actionable for us.

PR #1145, opened 2020-07-25
Cancel download should remove currently downloaded map

Ph0tonic:
I had a look at the code of the offline plugin and find out what might cause the cancel download not to delete the currently downloading map.

This is a PR to fix issue #189 

Note : I have not been able to actually test that this fix the issue as there is no cancel button in the demo app but in any case what I changed seems incorrect to me and might prevent the map to be deleted.

Any comment and review is most welcome @tobrun, @LukasPaczos and @langsmith


Ph0tonic:
I have finally been able to test it and it's working as expected. The problem happened only when the UUID of an `OfflineDownloadOptions` was bigger than an int and was converted to an int instead of keeping it as a long.

@LukasPaczos and @tobrun This PR is now definitely ready for a review 👍

Ph0tonic:
Any update about this ? Could we merge ?
\cc @LukasPaczos

Ph0tonic:
Any updates on this ?
\cc @tobrun & @LukasPaczos

Ph0tonic:
Hello, any updates on this ?
\cc @tobrun & @LukasPaczos

Ph0tonic:
Hello,
Any updates on this ?
@mapbox/maps-android

PR #1159, opened 2020-09-29
Prevent annotation changes from invalidating unrelated drag actions

### LINKED ISSUE #1029: Android animation cancels drag events on annotations
I'm using a CircleManager to handle a bunch of events such as click, long click, drag started, drag and drag finished. I am also using some android animations that do things such as periodically change the circles' radius.

Without the animations the drag events work just fine.

Now, with the animation process on, I can't drag my circles around. 
Using some traces I realized that the onAnnotationDragStarted event does get triggered when I try to move a circle. However, the onAnnotationDragFinished also instantly gets triggered right after the onAnnotationDragStarted.

Am I doing something work or is this expected behaviour ? If so, is there any workaround for it ? I'd very much like to be able to both animate and drag around my annotations.

Thank you

TroyStopera:
Relates to, but does not resolve, #1029 

### Motivation for change

I have a use case where markers on the map are 'linked' together via lines. As the user drags a marker, it is necessary that the line is programmatically updated as well. Due to the current logic which ensures data integrity, updates to the associated lines invalidate/stop the drag action.

This simple change adds a new method to the `DraggableAnnotationController` which allows the `AnnotationManager` to notify the controller of more fine-grain changes to the map source. If the change wasn't to the dragged annotation, the drag will continue.

### Future work

As stated above, this does not resolve #1029, which asks that changes to certain attributes of a dragged annotation don't stop the drag action. For example, a circle should be able to change size while being dragged, as this does not logically interfere with the annotation moving. 

Although not included in this PR, a solution to that could be to add a field/method to the `Annotation` class which computes a 'drag integrity hash' which hashes all fields that must remain static during a drag. When a drag starts, the `DraggableAnnotationManager` can take note of this hash. When notified of an update, the controller can check the current integrity hash of the annotation, and if it doesn't match the original stop the drag. 

If that solution is agreeable, I can update this PR to include that.

### DIFF
--- a/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
+++ b/plugin-annotation/src/main/java/com/mapbox/mapboxsdk/plugins/annotation/AnnotationManager.java
@@ -131,7 +131,7 @@ public abstract class AnnotationManager<
     T t = options.build(currentId, this);
     annotations.put(t.getId(), t);
     currentId++;
-    updateSource();
+    internalUpdateSource();
     return t;
   }
 
@@ -150,7 +150,7 @@ public abstract class AnnotationManager<
       annotations.put(annotation.getId(), annotation);
       currentId++;
     }
-    updateSource();
+    internalUpdateSource();
     return annotationList;
   }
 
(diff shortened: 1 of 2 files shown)


troy-skydio:
@LukasPaczos Note sure what I need to do to get eyes on this. I'd really like to get this working in master, so that I can update to the latest versions released by Mapbox.

LukasPaczos:
Tagging @mapbox/maps-android here.

tobrun (approved):

tobrun:
CI needs to be restored for this repo before we will be able to merge

PR #1164, opened 2020-11-14
Passes unhandled map interactions to lower annotations

TroyStopera:
When all `OnAnnotationLongClickListener` or `OnAnnotationClickListener` listeners return false, the event will be passed to the next annotation returned from the feature query.


troy-skydio:
@LukasPaczos Note sure what I need to do to get eyes on this. I'd really like to get this working in master, so that I can update to the latest versions released by Mapbox.

LukasPaczos:
Tagging @mapbox/maps-android here.

PR #1165, opened 2020-11-17
Improve MarkerView flick behavior

OttyLab:
## Background
MarkerView position is only updated with`onCameraDidChange`. When a user flicks the map, MarkerView keeps the position until the flick animation is finished as below. This behavior looks laggy from user point of view.
before

## Update
Call `update()` in `onCameraIsChanging` as well.

## Result
after

### DIFF
--- a/plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
+++ b/plugin-markerview/src/main/java/com/mapbox/mapboxsdk/plugins/markerview/MarkerViewManager.java
@@ -13,7 +13,7 @@ import androidx.annotation.UiThread;
 /**
  * Class responsible for synchronising views at a LatLng on top of a Map.
  */
-public class MarkerViewManager implements MapView.OnCameraDidChangeListener {
+public class MarkerViewManager implements MapView.OnCameraDidChangeListener, MapView.OnCameraIsChangingListener {
 
   private final MapView mapView;
   private final MapboxMap mapboxMap;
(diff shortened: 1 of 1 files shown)


Chaoba (commented):
Looks good. But what I concerned about is the performance. Currently, it has only 20 markers on the test Activity. Can you try to add more markers(for example 1000)  in `MarkerViewActivity` and see whether it will have a performance issue?

OttyLab:
@Chaoba Thanks! Sure, I'll do the test. BTW, do you have any metrics about performance? If no, then I'll just check if I can see a laggy behavior on some Android physical devices (Nexus 5 and Pixel 4a would be candidates)

Chaoba:
@OttyLab AFAIK, we don't have any metrics for makerview.  
Yes, please check on your device.

OttyLab:
@Chaoba 
Here's the result of physical device test. If the device is not powerful, this update makes the animation a bit laggy. But it's not critical, I suppose.

## Pixel 4a
### w/o update
#### 1000
o_1000

#### 10000
o_10000

### w/ update
#### 1000
u_1000

#### 10000
u_10000


## Nexus 7
### w/o update
#### 1000
no_1000

### w/ update
#### 1000
nu_1000

Chaoba (approved):
@OttyLab Thanks for your tests. LGTM
@tobrun Would you like to take a final look at this?

tobrun (approved):

g4rb4g3:
Could you guys please make a new release for MarkerViewPlugin too? The latest MarkerViewPlugin v0.4.0 release doesn't include this fix.

PR #1194, opened 2021-08-16
Android 12 compartibility fixes.

Chaoba:
Included changes:

- [x]  LeacyCanary, update to 2.7, no need to config for release and test. No need any config in codes.
- [x]  Multidex, methods num exceed limit after bump all the dependency. 
- [x] Add `android:exported="true"` in manifest and config for PendingIntent for Android 12
- [x] Bump maps sdk to 9.6.2 which contains the telemetry fix for Android 12

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -12,6 +12,7 @@ android {
     applicationId "com.mapbox.mapboxsdk.plugins.testapp"
     testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
     vectorDrawables.useSupportLibrary = true
+    multiDexEnabled = true
   }
 
   buildTypes {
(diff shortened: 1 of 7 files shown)


kiryldz (commented):
@Chaoba could you update PR description why actually leak canary was dropped, several dependencies were added, multidex was enabled. Is it all connected with updating to Android 12?

Chaoba:
A separate [ticket](https://github.com/mapbox/mapbox-plugins-android/issues/1196) is created to track the failed release job.

kiryldz:
@Chaoba is it possible to separate this PR in 2 PRs:
1. Android 12 compatibility
2. All other work done here.

I see no sense in messing it all up in one PR despite changes for leak canary, firebase etc are __indeed__ related to Android 12 compatibility.

app/build.gradle
```diff
@@ -64,12 +52,15 @@ dependencies {
     exclude group: 'android.arch.lifecycle'
     exclude group: 'com.google.android.gms'
   })
-  implementation dependenciesList.mapboxGeocoding
+  implementation (dependenciesList.mapboxGeocoding, {
+    exclude group: 'com.squareup.okhttp3'
+  })
```
pengdev, line 57: why do we exclude the okhttp specifically?

Chaoba: It will conflict with another version of okhttp and cause a compile error.

tobrun (approved):

PR #1202, opened 2021-09-10
Add locale for vi

Chaoba:
In some devices `Locale("vi", "VN")` is not supported for VIETNAM, this pr add `Locale("vi")` for these devices.


plugin-localization/src/main/java/com/mapbox/mapboxsdk/plugins/localization/MapLocale.java
```diff
@@ -348,6 +348,7 @@
     LOCALE_SET.put(new Locale("pt", "PT"), PORTUGAL);
     LOCALE_SET.put(new Locale("pt", "BR"), BRAZIL);
     LOCALE_SET.put(new Locale("vi", "VN"), VIETNAM);
+    LOCALE_SET.put(new Locale("vi"), VIETNAM);
```
kiryldz, line 351: How did you test it? Can we somehow make sure it indeed fixes an issue fully?

Chaoba: A customer already tested it with the snapshot.

kiryldz (approved):

PR #1205, opened 2021-10-04
Annotation order vs click priority inverted

felix-ht:
The order of annotations is inverted relative to the click priority. 
This causes annotations on lower (potentially occluded) layers, to priority the as opposed to the one actually visible to the user.

A detailed example of the issue can be found here:

https://github.com/tobrun/flutter-mapbox-gl/issues/563

Getting the last feature instead of the first one should fix this. 

**NOTE**
I tried testing this, but struggled to get a working setup - gradle causing build issues - so it would be fantastic is someone else could test this or help me getting a working test setup.


felix-ht:
This is unlikely to be the cause of the issues, as it only queries the features on the current layer

